PR #11, opened 2018-08-15
Fix for starting docker on slaves

tplavcic:
I noticed this issue when working for pt.cd.percona.com. Docker service was not started on the nodes.


delgod:
I think we can delete `.DS_Store` from commit

PR #14, opened 2018-08-28
Allow Parallel Execution of Stages for Different client tests

puneet0191:
Just trying this one out, not sure if the Architecture would allow this completely since we are not using different agents for each stage.


delgod:
have you checked? Is it works in parallel ok?

puneet0191:
I am not sure what's the best way to test this one directly without making merges, Ideally it should be able to run in Parallel okay, as I mentioned on the comment may be not having individual stage agents could be a cause of worry.

PR #20, opened 2018-11-13
JEN-952 change AZ

delgod:


IaC/ps.cd/JenkinsStack.yml
```diff
@@ -280,7 +280,7 @@ Resources:
   JSlaveRole: # create Role for jenkins slaves (needed for assign tags)
```
AlekSi: Should that comment be updated too?

delgod: good catch, will update later

delgod: fixed

PR #41, opened 2019-02-18
Replace ext2 with xfs on jenkins slaves for running PSMDB tests

tplavcic:
We have measured that ext2 is too slow for running PSMDB tests so we would like to change the mount option from ext2 to xfs for running PSMDB tests.
I'll open a Jira ticket for this also.


hors:
It was done in https://github.com/Percona-Lab/jenkins-pipelines/commit/c6462eca5d727c0589be2f9f83bf06e998bc8d90#diff-361f9303c35df426d7d315722419b6fe So, I am closing this PR

PR #43, opened 2019-02-18
JEN-1039 use xfs instead of ext2 for psmdb.cd

hors:
* fix cloud.groovy script

### DIFF
--- a/psmdb/percona-server-for-mongodb-3.2-param.yml
+++ b/psmdb/percona-server-for-mongodb-3.2-param.yml
@@ -59,7 +59,7 @@
         name: PSM_BRANCH
         trim: false
     - string:
-        default: '4'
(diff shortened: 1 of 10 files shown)


IaC/psmdb.cd/init.groovy.d/cloud.groovy
```diff
@@ -56,7 +57,8 @@ initMap['docker'] = '''
 
     if ! mountpoint -q /mnt; then
         DEVICE=$(ls /dev/xvdd /dev/nvme1n1 | head -1)
-        sudo mkfs.ext4 ${DEVICE}
+        sudo yum -y install xfsprogs
```
delgod: don't run yum install before loop in line 65, you can get failed init script

PR #51, opened 2019-03-11
PMM-3622 pmm2-server-autobuild Job

puneet0191:


pmm/pmm2-server-autobuild.groovy
```diff
@@ -0,0 +1,176 @@
...
+    stages {
+        stage('Prepare') {
+            steps {
+                installDocker()
+
+                git poll: true, branch: GIT_BRANCH, url: 'http://github.com/Percona-Lab/pmm-submodules'
+                sh '''
+                    git reset --hard
+                    sudo git clean -xdf
+                    git submodule update --init --jobs 10
```
EvgeniyPatlan: there was an issue due to lfs enabled in git. Is it works correctly with such commands now?

AlekSi: @EvgeniyPatlan Can you update git and install git lfs globally? What the other change did (wget'ing lfs, etc.) is an ad-hoc solution, and we would like to remove it.

pmm/pmm2-server-autobuild.groovy
```diff
@@ -0,0 +1,176 @@
...
+    }
+    stages {
+        stage('Prepare') {
+            steps {
+                installDocker()
+
+                git poll: true, branch: GIT_BRANCH, url: 'http://github.com/Percona-Lab/pmm-submodules'
+                sh '''
+                    git reset --hard
+                    sudo git clean -xdf
```
AlekSi: Why `sudo`?

EvgeniyPatlan: as I remember it is inheritance from pmm-server-autobuild job

AlekSi: Ok, but why? 😄 It shouldn't be needed.

pmm/pmm2-server-autobuild.groovy
puneet0191: yes it seems to works, I tested with a Job

pmm/pmm2-server-autobuild.groovy
puneet0191: I will remove it and check if it will work, thanks

pmm/pmm2-server-autobuild.groovy
```diff
@@ -0,0 +1,176 @@
...
+                stash includes: 'uploadPath', name: 'uploadPath'
+                archiveArtifacts 'shortCommit'
+                slackSend channel: '#pmm-ci', color: '#FFFF00', message: "[${JOB_NAME}]: build started - ${BUILD_URL}"
+            }
+        }
+        stage('Build client source') {
+            steps {
+                sh '''
+                    sg docker -c "
+                        export pmm_version=$(cat VERSION)
```
AlekSi: Why lowercase? Is it different from `PMM_VERSION`?
And where `VERSION` file is created?

puneet0191: I think, It is just to use this as an artifact on different stages

EvgeniyPatlan: pmm_version variable is used to set version of pmm. and VERSION file is in pmm-submodules repo

AlekSi: So is it the same as `PMM_VERSION`? Can we use a consistent way of writing it then? Otherwise, it looks like it is a related but different thing.

AlekSi: Wait. Nevermind. :) It is `pmm_version` everywhere. I was confusing it with `PMM_RELEASE_VERSION`.

PR #66, opened 2019-03-21
PMM-3716 PMM2-API Tests Job

puneet0191:


pmm/pmm2-api-tests.groovy
```diff
@@ -0,0 +1,114 @@
...
+                    sudo mv go /usr/local
+                    export GOROOT=/usr/local/go
+                    export GOPATH=/usr/local/go
+                    export PATH=$GOPATH/bin:$GOROOT/bin:$PATH
+                    go env
+                    sudo ln -s \$(pwd -P) $GOPATH/src/pmm-api-tests
+                    sudo curl https://raw.githubusercontent.com/golang/dep/master/install.sh | sh
+                    cd $GOPATH/src/pmm-api-tests/
+                    dep ensure -v
+                    make
```
AlekSi: @puneet0191 @BupycHuk Let's dockerise our tests instead so they can be run with just `make run` or something without installing Go, etc.

AlekSi: https://jira.percona.com/browse/PMM-3719

pmm/pmm2-api-tests.groovy
```diff
@@ -0,0 +1,98 @@
...
+            }
+        }
+        stage('Run API Test') {
+            steps {
+                withCredentials([sshUserPrivateKey(credentialsId: 'aws-jenkins', keyFileVariable: 'KEY_PATH', passphraseVariable: '', usernameVariable: 'USER')]) {
+                    sh '''
+                        sudo yum -y install docker
+                        sudo usermod -aG docker ec2-user
+                        sudo service docker start
+                        sudo docker build -t pmm-api-tests .
```
hors: Why do you build docker image every time when you run the job? Maybe it is better to build it once and push it to docker hub (Percona Lab) .

AlekSi: IMO, that okay: we only put Go binaries there, that does not take much time, and they (test binaries) change often.

pmm/pmm2-api-tests.groovy
```diff
@@ -0,0 +1,98 @@
...
+            }
+        }
+        stage('Start staging') {
+            steps {
+                runStaging(DOCKER_VERSION, CLIENT_VERSION, '--addclient=ps,1')
+            }
+        }
+        stage('Sanity check') {
+            steps {
+                sh "curl --silent --insecure '${PMM_URL}/prometheus/targets' | grep localhost:9090"
```
AlekSi: It is better to use `/ping` API – it returns 200 if PostgreSQL, Prometheus, etc. are up.

pmm/pmm2-api-tests.groovy
```diff
@@ -0,0 +1,98 @@
...
+        disableConcurrentBuilds()
+    }
+    triggers {
+        upstream upstreamProjects: '', threshold: hudson.model.Result.SUCCESS
+    }
+
+    stages {
+        stage('Prepare') {
+            steps {
+                // clean up workspace and fetch pmm-qa repository
```
AlekSi: But that's not pmm-qa repo

pmm/pmm2-api-tests.groovy
```diff
@@ -0,0 +1,98 @@
...
+            }
+        }
+        stage('Sanity check') {
+            steps {
+                sh "curl --silent --insecure '${PMM_URL}/prometheus/targets' | grep localhost:9090"
+            }
+        }
+        stage('Sleep') {
+            steps {
+                sleep 100
```
AlekSi: Once `/ping` returns 200 (that could take a while, we should try it in loop), there is no longer a need for sleep.

AlekSi: Alternatively, we may install Docker, etc., and then check /ping – to do more things in parallel, and reduce total time.

pmm/pmm2-api-tests.groovy
puneet0191: it will be in the future, once we merge changes to api-tests

pmm/pmm2-api-tests.groovy
puneet0191: hmm, sure makes sense. I checked `ping` this morning, doesn't seem to work, comes back with an empty response, `http://server_url/ping` right?, comes back as `{}`

AlekSi: That's expected. What is important is status code – 200 if everything is ready.

pmm/pmm2-api-tests.groovy
AlekSi: But then we can update comment too. I don't think that is a good idea to have a comment in sync with _future_ code, not with the current one. :)

PR #74, opened 2019-04-04
PMM-3716 Update yum missing before installing client

puneet0191:


pmm/aws-staging-start.groovy
```diff
@@ -288,8 +288,10 @@ pipeline {
 
                             if [[ \$CLIENT_VERSION = dev-latest ]]; then
                                 sudo yum -y install https://repo.percona.com/yum/percona-release-latest.noarch.rpm
-                                sudo percona-release enable original testing 
+                                sudo percona-release enable original testing
```
EvgeniyPatlan: I would suggest adding  `sudo percona-release disable all` as by default we enable our main repo

puneet0191: how about I use `enable-only`, `sudo percona-release enable-only original testing`

rnovikovP: my suggestions there  use  "as much as possible  real user conditions" 
To get all problems user can have. 
I haven't  seen we  recommend in our Documentation  `disable all`   much

PR #101, opened 2019-05-10
JEN-1100 install cmake3 for Centos 6/7

hors:


hors:
https://pxb.cd.percona.com/view/PXB%208.0/job/percona-xtrabackup-8.0-param-medium-test/3/

PR #107, opened 2019-05-18
JEN-1102 use rhel8 GA for rpms build

hors:


EvgeniyPatlan:
@hors please do the same for pmm2-client-autobuild

PR #121, opened 2019-05-31
PMM-4132 adding owner tag, and updating steps

puneet0191:


pmm/pmm2-api-tests.groovy
```diff
@@ -90,9 +94,9 @@ pipeline {
         always {
             script {
                 if (currentBuild.result == 'SUCCESS') {
-                    slackSend channel: '#pmm-ci', color: '#00FF00', message: "[${JOB_NAME}]: build finished"
+                    slackSend channel: '#pmm-ci', color: '#00FF00', message: "[${JOB_NAME}]: build finished, owner: @${OWNER}"
```
AlekSi: The idea was to notify the latest committer only if tests failed to avoid notifications fatigue.

puneet0191: okay sure, I will remove this notification in case of passed tests.

PR #165, opened 2019-07-19
PMM-4309 Adding Pipefail to avoid false positives

puneet0191:
AMI builds show pass even though, they failed in the build step, it is because of the piping we are using. This should reflect the real status code and also change the build status


AlekSi:
@puneet0191 please merge if this PR is ok

PR #207, opened 2019-08-21
Improve testing reports

DexterHD:
Small fix. Moved sh script...

### DIFF
--- a/pmm/pmm2-api-tests.groovy
+++ b/pmm/pmm2-api-tests.groovy
@@ -86,13 +86,13 @@ pipeline {
             steps {
                 sh '''
                     sudo docker run -e PMM_SERVER_URL=\${PMM_URL} pmm-api-tests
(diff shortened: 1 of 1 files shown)


pmm/pmm2-api-tests.groovy
```diff
@@ -86,13 +86,13 @@ pipeline {
             steps {
                 sh '''
                     sudo docker run -e PMM_SERVER_URL=\${PMM_URL} pmm-api-tests
+                    sudo docker cp pmm-api-tests:/go/src/github.com/Percona-Lab/pmm-api-tests/pmm-api-tests-junit-report.xml ./pmm-api-tests-junit-report.xml
```
puneet0191: this will not be executed every time, so you might end up getting no results.

DexterHD: Yes, but we can get report only after tests runs. And in my current implementation, it doesn't work at all, because this script doesn't run :( Check it here: https://pmm.cd.percona.com/blue/organizations/jenkins/pmm2-api-tests/detail/pmm2-api-tests/822/pipeline/39

PR #212, opened 2019-08-22
PMM-4448 Run PMM Server update test.

AlekSi:


pmm/pmm2-api-tests.groovy
```diff
@@ -85,16 +85,17 @@ pipeline {
...
             }
         }
     }
     post {
         always {
             sh '''
+                sudo docker ps -a
                 sudo docker cp ${BUILD_TAG}:/go/src/github.com/Percona-Lab/pmm-api-tests/pmm-api-tests-junit-report.xml ./pmm-api-tests-junit-report.xml
-                ls
+                ls -al
```
puneet0191: Was this needed for debugging? Or is intentionally left out?

AlekSi: It is both needed for debugging and intentionally left out :) I have an idea why `*.xml` was required, that should help to debug that problem – but only after merging.

PR #221, opened 2019-09-04
PMM-4309 submodules cleanup

puneet0191:


pmm/aws-staging-start.groovy
```diff
@@ -369,6 +369,7 @@ pipeline {
                                         echo "export PATH=$PWD/pmm2-client-2.0.0/bin:$PATH" >> ~/.bash_profile
                                         source ~/.bash_profile
                                         pmm-admin --version
+                                        cat pmm-agent.log
```
AlekSi: This file does not exist at this point.

puneet0191: When installation is done via tar or in all cases?

AlekSi: I mean, there is nothing before that line that creates this file. pmm-agent itself does not creates it, it logs only to stderr.

puneet0191: oh yes, I realize what I did there, pushing fix.

PR #228, opened 2019-09-11
PMM-4661 add possibility to skip syncing of pmm2-client rpm to repo

hors:


vars/sync2Prod.groovy
```diff
@@ -1,4 +1,4 @@
-def call(String DESTINATION) {
+def call(String DESTINATION, String SYNC_PMM_CLIENT) {
```
AlekSi: Is there a boolean type?

puneet0191: I think it is

PR #349, opened 2020-01-22
PMM-5064 Job for Submodule rewind for FB branches. 

puneet0191:
1. Removed the trigger, so it can only be manually triggered
2. Removed the use of version, honestly, we don't use it hence not needed.

### DIFF
--- a/pmm/pmm2-rewind-submodules-fb.groovy
+++ b/pmm/pmm2-rewind-submodules-fb.groovy
@@ -1,0 +1,96 @@
+pipeline {
+    agent {
+        label 'micro-amazon'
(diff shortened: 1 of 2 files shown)


pmm/pmm2-rewind-submodules-fb.groovy
```diff
@@ -0,0 +1,82 @@
+pipeline {
+    agent {
+        label 'micro-amazon'
+    }
+    parameters {
+        string(
+            defaultValue: 'PMM-2.0',
```
AlekSi: I think we need to remove this default.

PR #356, opened 2020-02-14
PMM-5363 Adding Support for Alert Manager for UI tests

puneet0191:


pmm/pmm2-ui-tests.groovy
```diff
@@ -147,6 +147,7 @@ pipeline {
                 sauce('SauceLabsKey') {
                     sauceconnect(options: '', sauceConnectPath: '') {
                         sh """
+                            export INSTANCE_IP=${SERVER_IP}
```
hors: @puneet0191  maybe it is better to do it once in env section ... https://github.com/Percona-Lab/jenkins-pipelines/blob/2ff14c75f803c01c111f1ed97509da88f8988230/pmm/pmm2-ui-tests.groovy#L36 section ...

puneet0191: I have re-used existing env variable in tests, so no need to initiate it in Env. section also, it is automatically done now.

PR #361, opened 2020-02-24
PSMDB-575 Start to use 'm5.2xlarge' instance for psmdb tests

hors:
* fix 'min-centos-6' worker
    * fix 'min-bionic' worker

### DIFF
--- a/IaC/psmdb.cd/init.groovy.d/cloud.groovy
+++ b/IaC/psmdb.cd/init.groovy.d/cloud.groovy
@@ -30,7 +30,7 @@ imageMap['min-stretch-x64'] = 'ami-0b9cb6198bfca486d'
 imageMap['min-trusty-x64'] = 'ami-08fbb070'
 imageMap['min-xenial-x64'] = 'ami-ba602bc2'
 imageMap['min-cosmic-x64'] = 'ami-080cf36c2a905f2b8'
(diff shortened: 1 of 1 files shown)


IaC/psmdb.cd/init.groovy.d/cloud.groovy
```diff
@@ -166,35 +231,51 @@ initMap['min-jessie-x64'] = '''
...
-typeMap['min-bionic-x64'] = typeMap['docker-32gb']
-typeMap['psmdb'] = typeMap['docker-32gb']
-typeMap['psmdb-bionic'] = typeMap['docker-32gb']
+typeMap['micro-amazon']      = 't2.small'
+typeMap['docker']            = 'c4.xlarge'
+typeMap['docker-32gb']       = 'm5.2xlarge'
+typeMap['min-centos-7-x64']  = typeMap['docker-32gb']
+typeMap['fips-centos-7-x64'] = typeMap['docker-32gb']
+typeMap['min-artful-x64']    = typeMap['docker-32gb']
+typeMap['min-centos-6-x64']  = typeMap['docker']
```
delgod: why not 32gb?

hors: We can not use m5 instance because aws does not allow it via ec2 plugin for centos 6, so for centos 6 I am using old m4 with 32gb.

PR #373, opened 2020-03-18
CLOUD-525 Fix minikube

hors:


cloud/jenkins/psmdb_operator_minikube.groovy
```diff
@@ -154,7 +154,7 @@ pipeline {
 
                         sudo curl -Lo /usr/local/bin/minikube https://storage.googleapis.com/minikube/releases/latest/minikube-linux-amd64
                         sudo chmod +x /usr/local/bin/minikube
-                        minikube start --vm-driver=none --dns-domain=percona.com --memory=4096 --cpus=3 --kubernetes-version v1.14.8
+                        minikube start --force  --dns-domain=percona.com --memory=4096 --cpus=3 --kubernetes-version v1.14.8
```
hors: @delgod  maybe we need to have possibility to set it (--kubernetes-version v1.14.8) up via option ?

delgod: it will be really great

hors: It was done and I added kube ver to file using which we indicate that  tests  finished successfully and allowed the concurrent builds. So now we can run several parallel tests for different versions of  kubernetes.

PR #426, opened 2020-05-11
CLOUD-515 Remove volumes left

cap1984:


cloud/jenkins/psmdb_operator_lke_version.groovy
```diff
@@ -30,6 +30,7 @@ void CreateCluster(String CLUSTER_PREFIX) {
 
 void ShutdownCluster(String CLUSTER_PREFIX) {
     sh '''
+        linode-cli volumes list --json | jq '.[] | select( (.label | startswith("pvc")) and .linode_id == null) | .id' | xargs -I {} linode-cli volumes delete {}
```
hors, line 33: Does this command remove volumes for particular cluster or nt?   As you know we call ShutdownCluster for each cluster.

cap1984: Nope. The command removes all volumes assigned previously to any cluster. Since cluster deletion does not lead us volumes removal, we need to clear them up with extra command.
I've changed the order of commands just to be sure that we will remove more unconnected volumes.

PR #450, opened 2020-06-01
PXC-3285 Add centos6 and 'debian jessie' docker images for pxc

hors:


EvgeniyPatlan:
do we really need jessie. it will be EOL on 30th of June?

hors:
> do we really need jessie? it will be EOL on 30th of June.

I know about it and informed  Ramesh too. He should contact you to discuss it.  It is not so hard from my side to add possibility of building this image (change one symbol).

PR #465, opened 2020-06-12
Bug fix : MTR failure output extraction

rameshvs02:


rameshvs02:
> There is also a typo on line 56 ($2 ~ /sipped/ - that should be skipped)

done

PR #472, opened 2020-06-26
PXC-3286 : Updated MTR output process function

rameshvs02:
Issue:  
Jenkins pipeline job is not showing MTR test failures with different combinations. 
for eg: replication tests with different binlog format.

Solution : 
Updated MTR output process function to combine test suite and combinations to print the unique result. 
sample output
```
$ grep rpl.rpl_variables_stm_innodb junit.xml
	<testcase name="rpl.rpl_variables_stm_innodb-mix"><failure/></testcase>
	<testcase name="rpl.rpl_variables_stm_innodb-stmt"><failure/></testcase>
$

### CODE REVIEW

**Comment by rameshvs02** [comment]
> Please add the task number into commit message

done

File: pxc/local/test-binary-pxc [hunk 1]
```diff
@@ -47,7 +47,7 @@ function process_mtr_output {
 
   export LC_ALL=C
   egrep -a -o '[[:alpha:]]+(_[[:alnum:]]+)?(\.\/)?.*\[ .* \].*$' ${INFILE} \
-    | awk '{ if ($3 == "disabled" || $3 == "skipped") { print $1 " " $3 } else if (match($2, /^w[0-9]/)) { print $1 " " $4 " " $6 } else { print $1 " " $5 " " $7 } }' \
+    | awk '{ if ($3 == "disabled" || $3 == "skipped") { print $1 " " $3 } else if (match($2, /^w[0-9]/)) { print $1 " " $4 " " $6 } else { print $1 " " "-" substr($2,2,length($2)-2) $5 " " $7 } }' \
```
kamil-holubicki, line 50: We don't need similar change for 'else if' where we service the situation of parallel tests execution (multiple workers)?. MTR framework can output something like:
`[ 30%] rpl.rpl_ssl1 'row'                       w2 [ pass ]  10664`

Additionally I think it would be good to have a comment above this egrep/awk explaining what we have on input and what $1, $2, etc... can be for particular possible input cases.

rameshvs02: @kamil-holubicki similar change is not required for `else if` condition. Current change will collect all multiple worker testcases result status
```
$ grep rpl.rpl_ssl1 junit.xml
	<testcase name="rpl.rpl_ssl1-mix" time="10.692"></testcase>
	<testcase name="rpl.rpl_ssl1-row" time="11.127"></testcase>
	<testcase name="rpl.rpl_ssl1-stmt" time="11.521"></testcase>
$
```

Added comment to MTR output process function, please check it and let me know if there are any changes required.

rameshvs02:
> Please explain in the commit message why/how the update is needed.

Updated commit message. Added detailed info in PR

pxc/local/test-binary-pxc
kamil-holubicki: `echo rpl.rpl_stm_lcase_tblnames     [ disabled ]  Bug#20697533 2015-03-20 erlend Fails several times each day on all platforms 
| awk '{ if ($3 == "disabled" || $3 == "skipped") { print $1 " " $3 } else if (match($2, /^w[0-9]/)) { print $1 " " $4 " " $6 } else { print $1 " " "-" substr($2,2,length($2)-2) $5 " " $7 } }'`
produces:
`rpl.rpl_stm_lcase_tblnames disabled`
THIS IS OK

`echo rpl.rpl_stm_lcase_tblnames                     [ skipped ]  ndbcluster disabled \
| awk '{ if ($3 == "disabled" || $3 == "skipped") { print $1 " " $3 } else if (match($2, /^w[0-9]/)) { print $1 " " $4 " " $6 } else { print $1 " " "-" substr($2,2,length($2)-2) $5 " " $7 } }'`
produces
`rpl.rpl_stm_lcase_tblnames skipped`
THIS IS OK

`echo rpl.rpl_stm_lcase_tblnames             w6 [ pass ]    120 
| awk '{ if ($3 == "disabled" || $3 == "skipped") { print $1 " " $3 } else if (match($2, /^w[0-9]/)) { print $1 " " $4 " " $6 } else { print $1 " " "-" substr($2,2,length($2)-2) $5 " " $7 } }'`
produces
`rpl.rpl_stm_lcase_tblnames pass 120`
THIS IS OK

`echo rpl.rpl_stm_lcase_tblnames 'mix'   [ skipped ]  Doesn't support --binlog-format='mixed' 
| awk '{ if ($3 == "disabled" || $3 == "skipped") { print $1 " " $3 } else if (match($2, /^w[0-9]/)) { print $1 " " $4 " " $6 } else { print $1 " " "-" substr($2,2,length($2)-2) $5 " " $7 } }'`
produces
`bash: syntax error near unexpected token `('`
THIS IS WRONG

`echo rpl.rpl_stm_lcase_tblnames 'mix'         w5 [ pass ]   1361 
| awk '{ if ($3 == "disabled" || $3 == "skipped") { print $1 " " $3 } else if (match($2, /^w[0-9]/)) { print $1 " " $4 " " $6 } else { print $1 " " "-" substr($2,2,length($2)-2) $5 " " $7 } }'`
produces
`rpl.rpl_stm_lcase_tblnames -ipass 1361`
THIS IS WRONG

Am I doing something wrong?

rameshvs02: @kamil-holubicki  Replaced MTR output process function with MTR XML report option in 5.7 jobs. Please check the latest 5.7 MTR job and let me know if there are any changes required. 
https://pxc.cd.percona.com/job/pxc-5.7-param/40/

rameshvs02:
> LGTM. I understand that using junit.xml fixes original problem (different binlog formats vs vailures) and now what we see in Jenkins comes from junit.xml generated by mtr framework, right?

Yes, now the junit.xml file is generated through MTR framework.

PR #504, opened 2020-07-20
PXC-3363: Add lsof for PXC pipelines

Sudokamikaze:


Sudokamikaze:
There's a request to add `lsof` in comment section

PR #511, opened 2020-07-23
PXC-3369: Add boost-devel package

Sudokamikaze:
The `shared_ptr.hpp` being missing on system which causes failure on pipeline run for CentOS 8

### DIFF
--- a/pxc/docker/install-deps
+++ b/pxc/docker/install-deps
@@ -67,7 +67,7 @@ if [ -f /usr/bin/yum ]; then
     "
 
     if [[ ${RHVER} -eq 8 ]]; then
-        PKGLIST+=" lz4 perl-Memoize perl-open valgrind-devel libubsan rpcgen libtirpc-devel"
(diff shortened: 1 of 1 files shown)


pxc/docker/install-deps
```diff
@@ -63,7 +63,7 @@ if [ -f /usr/bin/yum ]; then
     libzstd-devel tzdata check-devel scons libgcrypt-devel libev-devel zlib-devel libaio-devel \
     ncurses-devel readline-devel pam-devel socat vim-common openssl perl-CPAN rsync libcurl-devel \
     python3 sysbench perl-DBD-mysql \
-    patchelf lsof \
+    patchelf lsof boost-devel \
```
hors, line 66: Did you test it for Centos7/Centos6 ? If it is needed only for Centos8 please add it only for it.

Sudokamikaze: As I can see from the logs, it's only needed for CentOS 8, but Debian family has a `libboost-all-dev` package listed in `PKGLIST`

PR #517, opened 2020-07-30
PXC-3364 : Make MTR workers configurable for param jobs

rameshvs02:
Issue : 
          Currently, PXC-57/80 MTR pipeline jobs do not have the option to set custom MTR workers

Solution:
          Added the following parameters to set custom MTR workers
          PARALLEL_RUN : This value is used for all test suites except for Galera specific suites.
          GALERA_PARALLEL_RUN : This value is used for the Galera specific test suites
          MTR_SUITES : mysql-test-run.pl suite names
          FULL_MTR : Run MTR suites based on variable MTR_SUITES if the value is `no`. Otherwise, the full MTR will be performed.

PXC-5.7 test run result: https://pxc.cd.percona.com/job/test-pxc-5.7-param
PXC-8.0 test run result: https://pxc.cd.percona.com/job/test-pxc-8.0-param
          
PS : Removed debian jessie from the OS list as it has reached EOL

### DIFF
--- a/pxc/docker/run-test57
+++ b/pxc/docker/run-test57
@@ -22,6 +22,10 @@ docker run --rm \
     export MTR_ARGS='${MTR_ARGS}'
     export MTR_REPEAT='${MTR_REPEAT}'
     export MTR_VAULT_TOKEN='${MTR_VAULT_TOKEN}'
+	export FULL_MTR='${FULL_MTR}'
+	export GALERA_PARALLEL_RUN='${GALERA_PARALLEL_RUN}'
+	export MTR_SUITES='${MTR_SUITES}'
+	export PARALLEL_RUN='${PARALLEL_RUN}'
 
     mkdir /tmp/results
     cp /tmp/pxc/results/*.tar.gz /tmp/results
--- a/pxc/docker/run-test57
+++ b/pxc/docker/run-test57
@@ -34,7 +34,7 @@ docker run --rm \
 
     mkdir /tmp/results/log
     rsync -a --prune-empty-dirs --include '*/' --include '*.err' --exclude '*' /tmp/results/PXC/mysql-test/var* /tmp/results/log
-    sudo tar -C /tmp/results/log -czf "pxc57-test-mtr_logs.tar.gz" var1 var2 var3 || true
+    sudo tar -C /tmp/results/log -czf "pxc57-test-mtr_logs.tar.gz" var1 var2 var3
 
     sudo mv ./pxc57-test-mtr_logs.tar.gz /tmp/pxc/results/
     sudo mv /tmp/results/*.xml /tmp/results/*.output  /tmp/pxc/results/
--- a/pxc/docker/run-test57
+++ b/pxc/docker/run-test57
@@ -34,7 +34,7 @@ docker run --rm \
 
     mkdir /tmp/results/log
     rsync -a --prune-empty-dirs --include '*/' --include '*.err' --exclude '*' /tmp/results/PXC/mysql-test/var* /tmp/results/log
-    sudo tar -C /tmp/results/log -czf "pxc57-test-mtr_logs.tar.gz" var1 var2 var3
+    sudo tar -C /tmp/results/log -czf "pxc57-test-mtr_logs.tar.gz" var1 var2 var3 || true
 
     sudo mv ./pxc57-test-mtr_logs.tar.gz /tmp/pxc/results/
     sudo mv /tmp/results/*.xml /tmp/results/*.output  /tmp/pxc/results/
--- a/pxc/docker/run-test57
+++ b/pxc/docker/run-test57
@@ -33,8 +33,8 @@ docker run --rm \
     bash -x /tmp/scripts/test-binary-pxc57 /tmp/results
 
     mkdir /tmp/results/log
-    rsync -a --prune-empty-dirs --include '*/' --include '*.err' --exclude '*' /tmp/results/PXC/mysql-test/var* /tmp/results/log
-    sudo tar -C /tmp/results/log -czf "pxc57-test-mtr_logs.tar.gz" var1 var2 var3 || true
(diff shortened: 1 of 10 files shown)


pxc/local/test-binary-pxc
```diff
@@ -96,17 +65,137 @@ if [[ "${DEFAULT_TESTING}" != "no" ]]; then
...
-        ${MTR_ARGS} \
-        --force \
-        --max-test-fail=0 \
-        --suite-timeout=9999 \
-        --testcase-timeout=${TESTCASE_TIMEOUT} \
-        | tee ${WORKDIR_ABS}/mtr.output || status=$?
-    process_mtr_output PXC "${WORKDIR_ABS}/mtr.output" "${WORKDIR_ABS}/junit.xml"
+    if [ ${FULL_MTR} == "no" ]; then
+        # galera suite
+        if echo "$MTR_SUITES" | grep -Ewq "galera"; then
```
dutow, line 70: This implementation will ignore any non default value (innodb, gr, x, main, ...) specified to the MTR_SUITES argument, which is a bad idea: as jenkins only report failures, this implies that the entire suite succeeded, while in truth it wasn't even ran.

Please make sure that anything specified by the user in these configuration values will be correctly used by the script.

rameshvs02: Updated MTR script to support non-default MTR suites. Custom configurations will be used correctly through MTR_ARGS.

pxc/docker/run-test57
```diff
@@ -22,15 +22,19 @@ docker run --rm \
     export MTR_ARGS='${MTR_ARGS}'
     export MTR_REPEAT='${MTR_REPEAT}'
     export MTR_VAULT_TOKEN='${MTR_VAULT_TOKEN}'
+	export FULL_MTR='${FULL_MTR}'
```
dutow, line 25: Please fix formatting issues: previous lines used 4 spaces for indentation, these now use tabs

pxc/docker/run-test57
```diff
@@ -22,15 +22,19 @@ docker run --rm \
...
     mkdir /tmp/results
     cp /tmp/pxc/results/*.tar.gz /tmp/results
 
     bash -x /tmp/scripts/test-binary-pxc57 /tmp/results
 
     mkdir /tmp/results/log
-    rsync -a --prune-empty-dirs --include '*/' --include '*.err' --exclude '*' /tmp/results/PXC/mysql-test/var /tmp/results/log
-    sudo tar -C /tmp/results/log -czf "pxc57-test-mtr_logs.tar.gz" var
+    rsync -a --prune-empty-dirs --include '*/' --include '*.err' --exclude '*' /tmp/results/PXC/mysql-test/mtr_var /tmp/results/log
+    sudo tar -C /tmp/results/log -czf "pxc57-test-mtr_logs.tar.gz" mtr_var || true
```
dutow, line 37: Why isn't it an error if we can't create an archive? Please add a comment about it, as it's not trivial

pxc/jenkins/pxc57-pipeline.groovy
```diff
@@ -23,7 +23,7 @@ pipeline {
             name: 'PXB24_BRANCH',
             trim: true)
         choice(
-            choices: 'centos:6\ncentos:7\ncentos:8\nubuntu:xenial\nubuntu:bionic\nubuntu:focal\ndebian:jessie\ndebian:stretch\ndebian:buster',
+            choices: 'centos:6\ncentos:7\ncentos:8\nubuntu:xenial\nubuntu:bionic\nubuntu:focal\ndebian:stretch\ndebian:buster',
```
dutow, line 26: This change (and a few similar, like pxb versions) is not related to the commit message. Are you sure we should include then in this PR?

pxc/local/test-binary-pxc
```diff
@@ -96,17 +65,58 @@ if [[ "${DEFAULT_TESTING}" != "no" ]]; then
...
-        --max-test-fail=0 \
-        --suite-timeout=9999 \
-        --testcase-timeout=${TESTCASE_TIMEOUT} \
-        | tee ${WORKDIR_ABS}/mtr.output || status=$?
-    process_mtr_output PXC "${WORKDIR_ABS}/mtr.output" "${WORKDIR_ABS}/junit.xml"
+    mkdir mtr_var
+    if [ ${FULL_MTR} == "no" ]; then
+        ARRAY_MTR_SUITES=($(echo $MTR_SUITES | sed 's/,/ /g'))
+        for suite in "${ARRAY_MTR_SUITES[@]}"; do
+	        if [[ "$suite" == *"galera"* ]]; then
```
dutow, line 72: What about galera_sr, galera_3nodes, galera_3nodes_sr, ... ?

dutow (commented):
I've added some minor comments.

Also, I still see one possible design issue with this: currently on 5.6 and 5.7, we already run the full mtr suite by default, and we plan to do the same for 8.0. And this diff won't help the galera suites in that configuration, only if we don't run the full mtr suite.
The default also looks to be the non full mtr, which is not what we want: by default we should run all suites, to ensure that we don't break anything.

pxc/docker/run-test57
rameshvs02: done

pxc/docker/run-test57
rameshvs02: It was added for testing purposes, removed `|| true` from archive creation.

pxc/jenkins/pxc57-pipeline.groovy
rameshvs02: Debian Jessie has reached EOL, so thought of removing it. I think we include this in the PR, will update the PR comment.

pxc/local/test-binary-pxc
rameshvs02: This condition will also work with other Galera suites.

dutow (approved):

PR #561, opened 2020-09-16
ENG-906: Fix installation of git on CentOS

Sudokamikaze:
Hrvoje reported that CentOS 8 doesn't have `git` installed:
With `|| : ` it only installs java package

Here's the sample:
```
++ sudo yum -y install java-1.8.0-openjdk git aws-cli

Last metadata expiration check: 0:00:03 ago on Wed 16 Sep 2020 04:44:46 PM UTC.
No match for argument: aws-cli
Error: Unable to find a match: aws-cli
++ :
<<>>
Dependencies resolved.
==========================================================================================
 Package                       Arch    Version                            Repo        Size
==========================================================================================
Installing:
 java-1.8.0-openjdk            x86_64  1:1.8.0.262.b10-0.el8_2            AppStream  323 k
Installing dependencies:
```

### DIFF
--- a/IaC/ps80.cd/init.groovy.d/cloud.groovy
+++ b/IaC/ps80.cd/init.groovy.d/cloud.groovy
@@ -187,7 +187,8 @@ initMap['micro-amazon'] = '''
         sleep 1
         echo try again
     done
-    sudo yum -y install java-1.8.0-openjdk git aws-cli || :
+    sudo yum -y install java-1.8.0-openjdk aws-cli || :
+    sudo yum -y install git || :
     sudo yum -y remove java-1.7.0-openjdk || :
     sudo install -o $(id -u -n) -g $(id -g -n) -d /mnt/jenkins
 '''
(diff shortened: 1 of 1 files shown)


EvgeniyPatlan (changes requested):
You should investigate why it is not possible to install aws-cli and fix

Sudokamikaze:
> You should investigate why it is not possible to install aws-cli and fix

It doesn't appear to be available in public repositories. It can be installed by official way: https://docs.aws.amazon.com/cli/latest/userguide/install-cliv2.html

If it's OK, I'll implement it

IaC/ps80.cd/init.groovy.d/cloud.groovy
```diff
@@ -187,7 +187,7 @@ initMap['micro-amazon'] = '''
         sleep 1
         echo try again
     done
-    sudo yum -y install java-1.8.0-openjdk git aws-cli || :
```
hors, line 190: It should work for micro-amazon because it is Amazon Linux 2.

hors (changes requested):

hors (approved):

EvgeniyPatlan (approved):

Sudokamikaze:
Jenkins ignores `|| true` so I had to move `git` out of one `yum install` in d9e93de 
/cc @EvgeniyPatlan @hors

PR #586, opened 2020-09-28
PMM-5804 Updating UI tests job, client instance setup

puneet0191:


Sudokamikaze (approved):
Syntax looks good for me

puneet0191:
I have tested it with temp job, will merge it when other changes are also ready thanks for review!

pmm/pmm2-ui-tests.groovy
```diff
@@ -132,51 +159,63 @@ pipeline {
...
-                sh """
-                curl --data '{"enable_stt": true, "enable_telemetry": true}' -u admin:admin -X POST ${PMM_UI_URL}/v1/Settings/Change
-                curl -u admin:admin -X POST ${PMM_UI_URL}/v1/management/SecurityChecks/Start
-                """
-                sleep 300
-                sh """
-                curl --data '{"disable_stt": true, "enable_telemetry": true}' -u admin:admin -X POST ${PMM_UI_URL}/v1/Settings/Change
-                """
+        stage('Setup') {
+            parallel {
```
Sudokamikaze, line 183: Is it really needed to run this block in parallel? For example, just in case if sanity check fails, the pipeline will run till it finishes the whole stage. 

Other parts is good to me

pmm/pmm2-ui-tests.groovy
```diff
@@ -132,51 +159,63 @@ pipeline {
...
                 expression { env.GIT_COMMIT_HASH.length()>0 }
             }
             steps {
                 sh 'git checkout ' + env.GIT_COMMIT_HASH
             }
         }
-        stage('Start staging') {
-            steps {
-                runStaging(DOCKER_VERSION, CLIENT_VERSION, '--addclient=ps,1 --addclient=mo,2 --with-replica --addclient=pgsql,1 --addclient=pxc,1 --with-proxysql --pmm2 --setup-alertmanager --add-annotation', CLIENT_INSTANCE, SERVER_IP)
+        stage('Start PMM Server') {
```
Sudokamikaze, line 169: Better to be `Start PMM Server Instance` as on line 177, there's: `Start PMM Client Instance`

puneet0191: sure will make that correction thanks.

pmm/pmm2-ui-tests.groovy
puneet0191: yes, but it will still be very rare as compared to the number of jobs which will save 2-3 mins from this in parallel, but if you think it is not needed, I can move sanity Check out of this parallel block

puneet0191 (commented):
just replied

pmm/pmm2-ui-tests.groovy
Sudokamikaze: Got it, LGTM

Sudokamikaze (approved):

PR #596, opened 2020-10-02
DISTMYSQL-58: Fix creation of repo on RHEL for PDPS/PDPXC

Sudokamikaze:


ps-pxc-dist/rpm_release.sh
```diff
@@ -60,7 +60,11 @@ for REPOPATH in $REPOPATH_TMP; do
         #
     done
     #
-    createrepo --update /srv/${REPOPATH}/${REPOCOMP}/${rhel}/SRPMS
-    createrepo --update /srv/${REPOPATH}/${REPOCOMP}/${rhel}/RPMS/${arch}/
+    for rhel in ${RHVERS}; do
+        for arch in $(ls -1 redhat/${rhel}); do
+            createrepo --update /srv/${REPOPATH}/${REPOCOMP}/${rhel}/SRPMS
+            createrepo --update /srv/${REPOPATH}/${REPOCOMP}/${rhel}/RPMS/${arch}/
```
EvgeniyPatlan, line 66: can be moved to line 58. so we would not need one more for loop

EvgeniyPatlan (changes requested):

EvgeniyPatlan (approved):

PR #634, opened 2020-11-09
ENG-935: Set JAVA_HOME before starting Jenkins service

Sudokamikaze:
Alternatives set a JAVA_HOME on its run, but to be extra sure, it's better to put actual, not symlinked one to the real path of JAVA_HOME into `/etc/profile.d` & `/etc/sysconfig/jenkins`

### DIFF
--- a/IaC/pg.cd/JenkinsStack.yml
+++ b/IaC/pg.cd/JenkinsStack.yml
@@ -613,6 +613,11 @@ Resources:
                   echo JENKINS_HOME=/mnt/$JENKINS_HOST \
                       | tee -a /etc/sysconfig/jenkins
                   sed -i 's^JENKINS_ARGS=""^JENKINS_ARGS="--sessionTimeout=10080 --sessionEviction=86400"^' /etc/sysconfig/jenkins
+                  JAVA_HOME=$(alternatives --display java | grep current | sed 's/link currently points to //' | sed 's|/bin/java||' | xargs)
+                  echo "$JAVA_HOME" > /etc/profile.d/java.sh
+                  echo JENKINS_JAVA_CMD=$JAVA_HOME/bin/java \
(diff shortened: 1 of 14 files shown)


IaC/pg.cd/JenkinsStack.yml
```diff
@@ -613,6 +613,11 @@ Resources:
                   echo JENKINS_HOME=/mnt/$JENKINS_HOST \
                       | tee -a /etc/sysconfig/jenkins
                   sed -i 's^JENKINS_ARGS=""^JENKINS_ARGS="--sessionTimeout=10080 --sessionEviction=86400"^' /etc/sysconfig/jenkins
+                  JAVA_HOME=$(alternatives --display java | grep current | sed 's/link currently points to //' | sed 's|/bin/java||' | xargs)
+                  echo "$JAVA_HOME" > /etc/profile.d/java.sh
+                  echo JENKINS_JAVA_CMD=$JAVA_HOME/bin/java \
+                      | tee -a /etc/sysconfig/jenkins
```
delgod, line 619: can we add JAVA_HOME also into `/etc/sysconfig/jenkins` ?

delgod (changes requested):

IaC/pg.cd/JenkinsStack.yml
Sudokamikaze: Nope, we can't
But the JENKINS_JAVA_CMD we can

hors (commented):
@Sudokamikaze Did you find the root of the issue ? Can you reproduce it? We need to know what we are trying to fix. I am 100% sure that it is some regression after the update. I saw it the first time when  you updated the jenkins farm(maybe firstly). It is critical to find root of the issue  and have STR before you try to fix it otherwise it is shot to the sky.

PR #636, opened 2020-11-09
PXC-3473 : Build Galera in debug mode

rameshvs02:


pxc/local/build-binary-galera3
```diff
@@ -55,6 +55,10 @@ fi
 # ------------------------------------------------------------------------------
 
 pushd ${SOURCEDIR}
-    scons
+    if [[ "${CMAKE_BUILD_TYPE}" = "Debug" ]]; then
+       scons debug=0
+    else
+       scons
```
dutow, line 61: Shouldn't we also specify psi=1 for both cases?

venkatesh-prasad-v: psi = 1 should be used only for libgalera_smm.so and not for garbd. garbd crashes if we built with psi=1.(https://jira.percona.com/browse/PXC-3428).

I think we should build libgalera_smm.so and garbd separetely with different arguments.

rameshvs02: Galera3 build script is only used for PXC-5.6 MTR job and psi=1 is not specified in PXC-5.6 build script. Please let me know if we need to add psi=1 for PXC-5.6 MTR job. 
https://github.com/percona/percona-xtradb-cluster/blob/5.6/build-ps/build-binary.sh#L338

pxc/local/build-binary-galera3
```diff
@@ -55,6 +55,10 @@ fi
 # ------------------------------------------------------------------------------
 
 pushd ${SOURCEDIR}
-    scons
+    if [[ "${CMAKE_BUILD_TYPE}" = "Debug" ]]; then
+       scons debug=0 --config=force
```
venkatesh-prasad-v, line 59: `--config=force` may not work properly if we are building unittests. Please check this once locally.
```
/usr/bin/ld: /home/venki/work/pxc/3353/percona-xtradb-cluster-galera/galerautils/tests/gu_utils_test.c:64: undefined reference to `_mark_point'
/usr/bin/ld: /home/venki/work/pxc/3353/percona-xtradb-cluster-galera/galerautils/tests/gu_utils_test.c:65: undefined reference to `_ck_assert_failed'
/usr/bin/ld: /home/venki/work/pxc/3353/percona-xtradb-cluster-galera/galerautils/tests/gu_utils_test.c:65: undefined reference to `_mark_point'
/usr/bin/ld: /home/venki/work/pxc/3353/percona-xtradb-cluster-galera/galerautils/tests/gu_utils_test.c:68: undefined reference to `_ck_assert_failed'
/usr/bin/ld: /home/venki/work/pxc/3353/percona-xtradb-cluster-galera/galerautils/tests/gu_utils_test.c:68: undefined reference to `_mark_point'
/usr/bin/ld: /home/venki/work/pxc/3353/percona-xtradb-cluster-galera/galerautils/tests/gu_utils_test.c:69: undefined reference to `_ck_assert_failed'
```

I guess it is probably better to just do
```
pushd ${SOURCEDIR}
    if [[ "${CMAKE_BUILD_TYPE}" = "Debug" ]]; then
       scons debug=0 --config=force libgalera_smm.so garb/garbd
    else
       scons --config=force libgalera_smm.so garb/garbd
    fi
popd
```

venkatesh-prasad-v (approved):

Sudokamikaze (approved):

PR #659, opened 2020-11-24
Switch back to m42xlarge for PSMDB.cd

Sudokamikaze:
The price became higher than 0.21, and both ps80 and psmdb.cd working fine with m42xlarge once again.


Sudokamikaze:
Moving to open from draft as there's no alerts from Psmdb.cd

delgod (approved):

PR #668, opened 2020-11-27
ENG-939: Copy last log before termination of Jenkins master instance

Sudokamikaze:
We've discussed this approach in release-team, which is most appropriate to save logs with reason to investigate after shutdown of instance

### DIFF
--- a/IaC/pg.cd/JenkinsStack.yml
+++ b/IaC/pg.cd/JenkinsStack.yml
@@ -616,7 +616,7 @@ Resources:
                   #echo "/usr/bin/aws s3 sync --sse-kms-key-id alias/jenkins-pmm-backup --sse aws:kms --exclude '*/caches/*' --exclude '*/config-history/nodes/*' --exclude '*/secretFiles/*' --delete /mnt/ s3://backup.cd.percona.com/" > /etc/cron.daily/jenkins-backup
                   #chmod 755 /etc/cron.daily/jenkins-backup
 
-                  printf "* * * * * root bash -c 'curl -s http://169.254.169.254/latest/meta-data/spot/instance-action | grep action && sh -c \"service jenkins stop; umount /mnt\" || :'\n* * * * * root sleep 30; bash -c 'curl -s http://169.254.169.254/latest/meta-data/spot/instance-action | grep action && sh -c \"service jenkins stop; umount /mnt\" || :'\n" > /etc/cron.d/terminate-check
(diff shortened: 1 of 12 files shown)


IaC/pg.cd/JenkinsStack.yml
```diff
@@ -616,7 +616,7 @@ Resources:
...
                   #chmod 755 /etc/cron.daily/jenkins-backup
 
-                  printf "* * * * * root bash -c 'curl -s http://169.254.169.254/latest/meta-data/spot/instance-action | grep action && sh -c \"service jenkins stop; umount /mnt\" || :'\n* * * * * root sleep 30; bash -c 'curl -s http://169.254.169.254/latest/meta-data/spot/instance-action | grep action && sh -c \"service jenkins stop; umount /mnt\" || :'\n" > /etc/cron.d/terminate-check
+                  printf "* * * * * root bash -c 'curl -s http://169.254.169.254/latest/meta-data/spot/instance-action | grep action && sh -c \"service jenkins stop; if [[ ! -d /mnt/logs-$(date +%d-%m-%Y) ]]; then mkdir /mnt/logs-$(date +%d-%m-%Y); fi; cp /var/log/jenkins/jenkins.log /mnt/logs-$(date +%d-%m-%Y); umount /mnt\" || :'\n* * * * * root sleep 30; bash -c 'curl -s http://169.254.169.254/latest/meta-data/spot/instance-action | grep action && sh -c \"service jenkins stop; if [[ ! -d /mnt/logs-$(date +%d-%m-%Y) ]]; then mkdir /mnt/logs-$(date +%d-%m-%Y); fi; cp /var/log/jenkins/jenkins.log /mnt/logs-$(date +%d-%m-%Y); umount /mnt\" || :'\n" > /etc/cron.d/terminate-check
```
EvgeniyPatlan, line 619: if spot would be stopped 2(or more) times in a day we will have only last log

EvgeniyPatlan (changes requested):

IaC/pg.cd/JenkinsStack.yml
Sudokamikaze: @EvgeniyPatlan Could you please check?

hors: @Sudokamikaze  @EvgeniyPatlan  but who will rotate these logs? Why do we need all these logs? There can be a few reasons why we have main server drop. It is a capacity issue in 98% of time and 1.6%  price and 0.5% that something is wrong with main jenkins server. If we are talking about the first two cases we do not really need  the log because we can detect this issue via aws console. If we are talking about 3rd case we need the log but I think we need to have only the latest. P.S. this log can be big so make sure that you will have enough time to back up it.

EvgeniyPatlan (approved):

delgod (approved):

EvgeniyPatlan (approved):

PR #685, opened 2020-12-09
Option to build PDF with MkDocs for PMM2

PaulJacobs-percona:
Added boolean to trigger PDF build (which takes 3-4x longer than HTML). PDF is only needed for releases so this option saves time when building HTML previews.


puneet0191 (approved):
looks good to me, have you tested it already once?

PaulJacobs-percona:
> looks good to me, have you tested it already once?

Yes, tested *thrice*, in use at https://pmm.cd.percona.com/job/pmm-2-x-documentation-mkdocs/

PR #687, opened 2020-12-09
Remove monitoring test in cloud tests for PSMDBO

tplavcic:
We don't support PMM 1.x any more and this test sometimes gets stuck so this will speed up release testing.


cap1984 (approved):

hors (approved):
Please do the same for PXCO

PR #704, opened 2020-12-23
ENG-960: Add repo.ci definition to hosts file of PSMDB.cd and REL.cd Jenkins

Sudokamikaze:
Currently deployed on REL.cd, the purpose is to remove infra/workers-related code from pipelines and move it to workers configruation

### DIFF
--- a/IaC/psmdb.cd/init.groovy.d/cloud.groovy
+++ b/IaC/psmdb.cd/init.groovy.d/cloud.groovy
@@ -78,9 +78,7 @@ initMap['docker'] = '''
         sudo mount -o noatime ${DEVICE} /mnt
     fi
 
-    cat /etc/hosts > ./hosts
-    echo '10.30.6.9 repo.ci.percona.com' >> ./hosts
-    sudo cp ./hosts /etc/
+    echo '10.30.6.9 repo.ci.percona.com' | sudo tee -a /etc/hosts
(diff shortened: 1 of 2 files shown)


IaC/psmdb.cd/init.groovy.d/cloud.groovy
```diff
@@ -78,6 +78,10 @@ initMap['docker'] = '''
         sudo mount -o noatime ${DEVICE} /mnt
     fi
 
+    cat /etc/hosts > ./hosts
+    echo '10.30.6.9 repo.ci.percona.com' >> ./hosts
+    sudo cp ./hosts /etc/
```
delgod, line 83: sometimes, `/etc/hosts` controlled externaly, so it is bad idea to recreate it
it is simpler to edit it in place to avoid issues
```suggestion
    echo '10.30.6.9 repo.ci.percona.com' | sudo tee -a /etc/hosts
```

delgod (changes requested):

IaC/psmdb.cd/init.groovy.d/cloud.groovy
Sudokamikaze: Tested on rel and reworked 713df27

delgod (approved):

PR #729, opened 2021-01-09
PMM-7252 fix dockerhub permissions

atymchuk:
[PMM-7252 ](https://jira.percona.com/browse/PMM-7252)
Pls refer to the ticket for the details.

### DIFF
--- a/pmm/pmm2-server-release.groovy
+++ b/pmm/pmm2-server-release.groovy
@@ -114,6 +114,7 @@ pipeline {
                         echo ${VERSION} > VERSION
                     """
                     sh '''
+                        set -ex
                         export VERSION=$(cat VERSION)
                         export TOP_VER=$(cat VERSION | cut -d. -f1)
                         export MID_VER=$(cat VERSION | cut -d. -f2)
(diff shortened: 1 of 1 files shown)


pmm/pmm2-server-release.groovy
```diff
@@ -159,7 +160,7 @@ pipeline {
                 withCredentials([usernamePassword(credentialsId: 'hub.docker.com', passwordVariable: 'PASS', usernameVariable: 'USER')]) {
                     sh """
                         sg docker -c "
-                            docker login -u "${USER}" -p "${PASS}"
+                            echo "${PASS}" | docker login -u "${USER}" --password-stdin
```
EvgeniyPatlan, line 163: would not it print $PASS into log? if yes it is security hole!

EvgeniyPatlan (changes requested):

pmm/pmm2-server-release.groovy
EvgeniyPatlan: Another question - why it worked correctly for pmm-server but failed to push pmm-client?

EvgeniyPatlan (changes requested):

pmm/pmm2-server-release.groovy
atymchuk: > Another question - why it worked correctly for pmm-server but failed to push pmm-client?

Because, one and the same account may have different permissions on different repositories. This was our case, and @edux helped resolve it.

atymchuk: > would not it print $PASS into log? if yes it is security hole!

>  Using STDIN prevents the password from ending up in the shell’s history, or log-files.

It's quote from here https://docs.docker.com/engine/reference/commandline/login/#provide-a-password-using-stdin

EvgeniyPatlan (approved):

puneet0191 (approved):

PR #747, opened 2021-01-20
Stash specifics; delete before unstashing

PaulJacobs-percona:
Changes mainly to eliminate doubt as to why website not being immediately refreshed after rsync.
Changed:
- Remove chmods (I guess not needed as this is just a simple copy of checked out git branch)
- Delete dir before unstashing: Stash was including this repo's files and other miscellaneous rubbish
- ls to see what's actually unstashed
- Stash of specific dirs/files (limits flexibility but will do until this job can be recreated in GHA)
- Change copied path (so rsync deletes EVERYTHING, not just what we copy) - Discovered after looking at web server folder and discovering historical remnants of other rsync's

### DIFF
--- a/pmm/pmm-documentation.groovy
+++ b/pmm/pmm-documentation.groovy
@@ -34,7 +34,7 @@ pipeline {
                         echo BRANCH=${BRANCH_NAME}
                         ls -l
                         DEST_HOST='docs-rsync-endpoint.int.percona.com'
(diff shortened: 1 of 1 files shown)


PaulJacobs-percona:
Turns out I've been using this PR for months -- consider it extended testing -- so let's use it instead of what's in main.

atymchuk (approved):

puneet0191 (approved):

PR #749, opened 2021-01-20
PMM-7352 Update AWS-Staging Job to use Library for spot Instance

puneet0191:
Tested the changes here: https://pmm.cd.percona.com/view/all/job/aws-staging-test-job/503/console

### DIFF
--- a/pmm/aws-staging-start.groovy
+++ b/pmm/aws-staging-start.groovy
@@ -3,6 +3,11 @@ import hudson.slaves.*
 import jenkins.model.Jenkins
 import hudson.plugins.sshslaves.SSHLauncher
 
+library changelog: false, identifier: 'lib@master', retriever: modernSCM([
+    $class: 'GitSCMSource',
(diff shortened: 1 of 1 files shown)


puneet0191:
We Will only merge this when we merge the PHP script also to pmm-qa, it has to go together to avoid failures.

atymchuk:
> We Will only merge this when we merge the PHP script also to pmm-qa, it has to go together to avoid failures.

I suggest that one be merged first.

puneet0191:
> > We Will only merge this when we merge the PHP script also to pmm-qa, it has to go together to avoid failures.
> 
> I suggest that one be merged first.

yes, it will break aws-staging hence we should do it within a quick time, will merge that first and this

puneet0191:
@atymchuk this is the other PR to add that file https://github.com/percona/pmm-qa/pull/377

atymchuk (approved):

PR #794, opened 2021-02-11
Remove rs-shard-migration test for PSMDBO minikube

tplavcic:
This test cannot work because of antiAffinity on minikube.


delgod (approved):

delgod (changes requested):
the subject is about minikube but the change in GKE pipeline

tplavcic:
@delgod nice catch! fixed!

cap1984 (approved):

delgod (approved):

PR #807, opened 2021-02-18
DISTPG-171. Add centos8.3 to list of images to test

mu-samoylov:


vars/ppgOperatingSystems.groovy
```diff
@@ -1,3 +1,3 @@
 def call() {
-  return ['centos-7', 'debian-9', 'debian-10', 'ubuntu-focal', 'ubuntu-bionic', 'rhel8']
+  return ['centos-7', 'centos8', 'debian-9', 'debian-10', 'ubuntu-focal', 'ubuntu-bionic', 'rhel8']
```
hors, line 2: Are you sure that we need to run the tests on  centos8 and rhel8?

mu-samoylov: Yes, we need that for different minor versions (8.2 and 8.3).

hors: Are you sure that we need to test 8.2? @EvgeniyPatlan  what do you think ?

EvgeniyPatlan: @mu-samoylov we need to perform tests always on latest available OS
otherwise we would need to add all OS from 7.0 to 7.9 and from 8.0 to 8.3

mu-samoylov: @EvgeniyPatlan are you sure that it's ok to say to customers update your distro to latest version? I don't think so. At current moment I prefer to have a both versions. Testing is in my area of responsibility, not yours. So please don't argue, but accept the pull request.

EvgeniyPatlan: yes

EvgeniyPatlan: at least please add comment which would describe what version for which image is used. Otherwise this looks very strange

mu-samoylov: Comment added.

EvgeniyPatlan (approved):
let it be

hors (approved):

PR #855, opened 2021-03-22
PMM-7490 - External Exporter 

IvanaBrecherova:


puneet0191 (approved):

puneet0191:
@Sudokamikaze please confirm if we using right credential id

Sudokamikaze (approved):
LGTM, checked with Jenkins, uses correct cred id

PR #872, opened 2021-03-29
ENG-993: Implement automatic fetch of SSH keys for our Jenkins farm

Sudokamikaze:


IaC/pmm.cd/JenkinsStack.yml
```diff
@@ -810,6 +830,21 @@ Resources:
                 systemctl reload sshd
               }
 
+              setup_ssh_keys() {
+                KEYS_LIST="alexander.tymchuk mykola.marzhan evgeniy.patlan slava.sarzhan illia.pshonkin alex.miroshnychenko eduardo.casarero santiago.ruiz andrew.siemen kamil.babayev"
+
+                for KEY in $KEYS_LIST; do
+                    if [[ ! -f "/tmp/$KEY.pub" ]]; then
+                        until wget -O "/tmp/$KEY.pub" "https://www.percona.com/get/engineer/KEY/$KEY.pub"; do
```
atymchuk, line 838: I think we should let it fail in case a person left the company and you are not aware or have not updated this file.

Sudokamikaze: We must not remove this conditional check

But you're right, we should improve it, if it returns 404, it should be skipped.

IaC/cloud.cd/JenkinsStack.yml
```diff
@@ -742,6 +762,26 @@ Resources:
...
+                            until wget -O "/tmp/$KEY.pub" "https://www.percona.com/get/engineer/KEY/$KEY.pub"; do
+                                echo "sleep before retry"
+                                sleep 1
+                            done
+                        elif [[ ${STATUS} -eq 404 ]]; then
+                            echo "Skipping key $KEY"
+                        fi
+                    fi
+                    cat /tmp/$KEY.pub | tee -a /home/ec2-user/.ssh/authorized_keys
+                    rm /tmp/$KEY.pub
```
delgod, line 781: I think these commands will fail if `${STATUS} -eq 404`
we have `set -o errexit` above

delgod (changes requested):

IaC/cloud.cd/JenkinsStack.yml
Sudokamikaze: You're right, I'll add flag

Sudokamikaze: Reworked in 01bac22

delgod (approved):

PR #874, opened 2021-03-29
PXC-3622: Implement JUnit output feature for 5.7

Sudokamikaze:
Param job URL: https://pxc.cd.percona.com/job/pxc-5.7-param-junittest/3/

Requires a rebuild of pxc-build docker image, should be performed after merge immediately.


EvgeniyPatlan (approved):
LGTM but please wait for other reviewers

venkatesh-prasad-v (approved):
Looks good to me

kamil-holubicki:
Just a question: does it solve the following problem:
MTR framework works in the way that it tries to rerun tests that fail. So if the test result is [ fail ] for the first time, MTR reurns it and the result may be [ retry-fail ] or [ retry-pass ].
In case of [ retry-pass ] will the test still be reported as failed?

kamil-holubicki (approved):
lgtm

PR #877, opened 2021-03-30
Adding Mailosaur Env Variables

puneet0191:


yurkovychv (approved):
Totally forgot about nightly job. Thanks a lot

PR #902, opened 2021-04-13
ENG-1001: Fix installation of python for RHEL in pxc57-build job

Sudokamikaze:
URL: https://pxc.cd.percona.com/view/PXC%205.7/job/pxc57-build/88/


pxc/pxc57-build.yml
```diff
@@ -102,6 +102,12 @@
 
         find ./target # For testing results and log
 
+        if [ -f /usr/bin/yum ]; then
+            if [[ $(rpm --eval %rhel) -eq 7 ]]; then
+                export PATH="/usr/local/bin:$PATH"
+            fi
+        fi
```
EvgeniyPatlan, line 109: why is it needed?

Sudokamikaze: Because CentOS 7's `python3-pip` installs awscli in /usr/local/bin which is not in default path.

EvgeniyPatlan (approved):

PR #917, opened 2021-04-21
PS-7650: Install AWSCLI for CentOS platforms as well

Sudokamikaze:
Needed for fips trunk

### DIFF
--- a/IaC/ps57.cd/init.groovy.d/cloud.groovy
+++ b/IaC/ps57.cd/init.groovy.d/cloud.groovy
@@ -131,8 +131,9 @@ initMap['docker'] = '''
     sudo yum -y remove java-1.7.0-openjdk awscli
 
     if ! $(aws --version | grep -q 'aws-cli/2'); then
-        find /tmp -maxdepth 1 -name "*aws*" | xargs sudo rm -rf
-
+        if [ -d /tmp/aws ]; then
+            sudo rm -rf /tmp/aws
+        fi
         until curl "https://awscli.amazonaws.com/awscli-exe-linux-x86_64.zip" -o "/tmp/awscliv2.zip"; do
             sleep 1
             echo try again
@@ -194,7 +195,9 @@ initMap['micro-amazon'] = '''
     sudo yum -y remove java-1.7.0-openjdk aws-cli || :
 
     if ! $(aws --version | grep -q 'aws-cli/2'); then
-        find /tmp -maxdepth 1 -name "*aws*" | xargs sudo rm -rf
+        if [ -d /tmp/aws ]; then
+            sudo rm -rf /tmp/aws
+        fi
 
         until curl "https://awscli.amazonaws.com/awscli-exe-linux-x86_64.zip" -o "/tmp/awscliv2.zip"; do
             sleep 1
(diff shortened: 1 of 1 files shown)


IaC/ps57.cd/init.groovy.d/cloud.groovy
```diff
@@ -180,13 +180,31 @@ initMap['micro-amazon'] = '''
...
+
+    if ! $(aws --version | grep -q 'aws-cli/2'); then
+        find /tmp -maxdepth 1 -name "*aws*" | xargs sudo rm -rf
+
+        until curl "https://awscli.amazonaws.com/awscli-exe-linux-x86_64.zip" -o "/tmp/awscliv2.zip"; do
+            sleep 1
+            echo try again
+        done
+
+        7za -o/tmp x /tmp/awscliv2.zip
```
EvgeniyPatlan, line 204: it will fail if aws dir is available

EvgeniyPatlan (changes requested):

IaC/ps57.cd/init.groovy.d/cloud.groovy
Sudokamikaze: It won't: https://github.com/Percona-Lab/jenkins-pipelines/pull/917/files#diff-b6ffba51adae65104db98666c6d5446fff1b01cfc9e891d410d67f13aa332e4cR197

IaC/ps57.cd/init.groovy.d/cloud.groovy
```diff
@@ -180,13 +180,31 @@ initMap['micro-amazon'] = '''
...
+    if [[ $SYSREL -eq 2 ]]; then
+        sudo amazon-linux-extras install epel -y
+    else
+        sudo yum -y install epel-release
+    fi
+    sudo yum -y install java-1.8.0-openjdk git p7zip || :
+    sudo yum -y remove java-1.7.0-openjdk aws-cli || :
+
+    if ! $(aws --version | grep -q 'aws-cli/2'); then
+        find /tmp -maxdepth 1 -name "*aws*" | xargs sudo rm -rf
```
EvgeniyPatlan, line 197: why not just sudo rm -rf /tmp/*aws* ?

EvgeniyPatlan (changes requested):

IaC/ps57.cd/init.groovy.d/cloud.groovy
Sudokamikaze: Because we rm -rf that dir only if it exists, usually it doesn't. 

All Jenkins have this command https://github.com/Percona-Lab/jenkins-pipelines/blob/master/IaC/rel.cd/init.groovy.d/cloud.groovy#L135

EvgeniyPatlan: well in case there is symlink or file which will match to *aws* it also will be removed.  And you need to remove aws dir only without * pattern
So why not 
if [ -d /tmp/aws ]; then
  sudo rm -rf /tmp/aws
fi

it will return only needed directory and will not touch other files/directories

Sudokamikaze: Ok, I will update it, and on next update of Jenkins, I'll replace that command with block you've suggested

EvgeniyPatlan (approved):

PR #952, opened 2021-05-07
Add Choice for AWS-Staging-start job which will trigger the install script

VoK0:
this add Choice for AWS-Staging-start job which will trigger the install script more specifically this one : install_backup_toolkit.sh


puneet0191:
This may no longer be required, hence closing this PR, can be directly setup using pmm-framework

PR #954, opened 2021-05-07
ENG-1011: Increase disk size of PS80 workers for package-testing job

Sudokamikaze:
Deployed on ps80


Sudokamikaze:
WIP, centos-6 ignores disk settings
Setting as draft

Sudokamikaze:
Due to:
http://lists.openstack.org/pipermail/openstack/2014-August/008721.html
Growpart called by cloud-init only works for kernels >3.8. Only newer kernels support changing the partition size of a mounted partition. When using an older kernel the resizing of the root partition happens in the initrd stage before the root partition is mounted and the subsequent cloud-init growpart run is a no-op.

I had to make a modified CentOS 6 image, with growpart utility installed and grew root partition to 30 GB manually

EvgeniyPatlan (approved):

PR #996, opened 2021-05-24
PMM-7993 use multiple client nodes to avoid client run out of memory

puneet0191:
Need to be able to set multiple client instances with different DB combinations as part of our Nightly job setup, currently, the instance runs out of memory in 30 mins resulting in tests failures and blocking us from adding more tests scenarios to the nightly job, the setup was already tested last week. 
https://pmm.cd.percona.com/blue/organizations/jenkins/pmm2-ui-tests-nightly-temp/detail/pmm2-ui-tests-nightly-temp/31/pipeline (ignore failure on UI tests)


Sudokamikaze (approved):
LGTM

PR #1029, opened 2021-06-07
PMM-8151 Update PMM UI Upgrade job to use Docker Compose setup with remote instances change

puneet0191:
1) No More needs two workers to execute upgrade tests.
2) Uses docker-compose setup for remote instances, which allows us more freedom to control test execution and set up the infra on the same worker. 

All the environment variables related to remote instances would be removed 2 weeks after we know the stability of the setup. 

The changes have been tested on temp jobs, https://pmm.cd.percona.com/blue/organizations/jenkins/pmm2-upgrade-tests-temp/detail/pmm2-upgrade-tests-temp/27/pipeline


pmm/pmm2-ui-tests.groovy
```diff
@@ -190,6 +190,9 @@ pipeline {
                     }
                     steps {
                         sh """
+                            PWD=\$(pwd) MYSQL_IMAGE=\${MYSQL_IMAGE} MONGO_IMAGE=\${MONGO_IMAGE} POSTGRES_IMAGE=\${POSTGRES_IMAGE} PMM_SERVER_IMAGE=\${DOCKER_VERSION} docker-compose up -d
+                            sleep 30
```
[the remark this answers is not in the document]
puneet0191: sure I have changed it, waiting for DB's to be ready, please re-check again.

Sudokamikaze (approved):

pmm/pmm2-ui-tests.groovy
```diff
@@ -190,8 +190,16 @@ pipeline {
                     }
                     steps {
                         sh """
-                            mkdir testdata
-                            sudo MYSQL_IMAGE=\${MYSQL_IMAGE} MONGO_IMAGE=\${MONGO_IMAGE} POSTGRES_IMAGE=\${POSTGRES_IMAGE} PMM_SERVER_IMAGE=\${DOCKER_VERSION} docker-compose up -d
+                            docker volume create pmm-server-data
+                            docker network create -d bridge --subnet 192.168.0.0/24 --gateway 192.168.0.1 pmm-network
```
Sudokamikaze, line 194: I think it's better to declare volumes and networks inside compose stack as they won't be removed on `docker-compose --down`

pmm/pmm2-ami-staging-start.groovy
```diff
@@ -175,7 +187,11 @@ pipeline {
                             set -o errexit
                             set -o xtrace
                             [ ! -d "/home/centos" ] && echo "Home directory for centos user does not exist"
-                            sudo yum -y install git svn sysbench
+                            sudo yum -y install git svn docker
+                            sudo systemctl start docker
+                            sudo curl -L https://github.com/docker/compose/releases/download/1.29.0/docker-compose-`uname -s`-`uname -m` | sudo tee /usr/bin/docker-compose > /dev/null
```
EvgeniyPatlan, line 192: it is not good idea to download and execute any binaries without at least checking md5

pmm/pmm2-upgrade-tests.groovy
```diff
@@ -2,114 +2,52 @@ library changelog: false, identifier: 'lib@master', retriever: modernSCM([
...
-                    ${USER}@${VM_IP}:workspace/aws-staging-start/pmm-agent.log \
-                    pmm-agent.log
-            fi
-        """
-    }
+    sh """
+        export CLIENT_VERSION=${CLIENT_VERSION}
+        if [[ \$CLIENT_VERSION != http* ]]; then
+            journalctl -u pmm-agent.service > /var/log/pmm-agent.log
+            sudo chmod 777 /var/log/pmm-agent.log
```
EvgeniyPatlan, line 40: why do you need such insecure permissions?

EvgeniyPatlan (changes requested):

pmm/pmm2-ui-tests.groovy
puneet0191: sure, I have addressed this, thanks for the input.

pmm/pmm2-ami-staging-start.groovy
puneet0191: I have added the check.

pmm/pmm2-upgrade-tests.groovy
puneet0191: yes, changed it, thanks a lot for the input.

yurkovychv (approved):

vars/waitForContainer.groovy
```diff
@@ -0,0 +1,19 @@
...
+        while [ \$attempt -le 30 ]; do
+            attempt=\$(( \$attempt + 1 ))
+            echo "Waiting for ${CONTAINER_NAME} to be up (attempt: \$attempt)..."
+            result=\$(docker logs ${CONTAINER_NAME} 2>&1)
+            if grep -q "${LOGS}" <<< \$result ; then
+              echo "${CONTAINER_NAME} is ready!"
+              break
+            fi
+            sleep 1
+        done;
```
EvgeniyPatlan, line 17: Do we really need this loop to be executed case 30 times and printed 30 times message?
If we know that running containers took up to 30 seconds why not just sleep for 30 seconds and then check logs
Or at least lets make interval for 10 seconds. So the loop will be executed only 3 times.

EvgeniyPatlan (changes requested):

vars/waitForContainer.groovy
puneet0191: good point, I will change it to run with 10 seconds wait, I think it was done initially with 30 seconds of sleep, but Illia suggested improving it with container startup wait

puneet0191: Done

vars/waitForContainer.groovy
```diff
@@ -0,0 +1,19 @@
+def call(String CONTAINER_NAME, String LOGS) {
+    sh """
+        set -o errexit
+        set -o xtrace
+        export CONTAINER_NAME=${CONTAINER_NAME}
+        export LOGS="${LOGS}"
+        attempt=0
+        while [ \$attempt -le 30 ]; do
+            attempt=\$(( \$attempt + 1 ))
```
EvgeniyPatlan, line 9: With change to `sleep 10` this loop will be executed for 300 seconds.So I suggest setting `\$attempt -le 4`

EvgeniyPatlan (changes requested):

vars/waitForContainer.groovy
puneet0191: yes, missed this, addressed, please check

EvgeniyPatlan (approved):

Sudokamikaze (approved):

yurkovychv (approved):

PR #1046, opened 2021-06-11
ENG-733: Update pxc.cd to 2.289.1

Sudokamikaze:
Update was deployed

ENG-1023 commit:
I've added corresponding labels for each worker, to decrease billing, I've made micro-amazon & docker multiuse
Tests can be demonstrated on DS

### DIFF
--- a/IaC/pxc.cd/init.groovy.d/cloud.groovy
+++ b/IaC/pxc.cd/init.groovy.d/cloud.groovy
@@ -23,16 +23,15 @@ netMap['us-west-1b'] = 'subnet-01d9a7d6b4722eb43'
 netMap['us-west-1c'] = 'subnet-0550c1d2ffd688021'
 
 imageMap = [:]
-imageMap['micro-amazon']     = 'ami-0bdb828fd58c52235'
-imageMap['micro-amazon2']    = 'ami-0577b787189839998'
-imageMap['min-bionic-x64']   = 'ami-07b068f843ec78e72'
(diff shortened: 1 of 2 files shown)


IaC/pxc.cd/init.groovy.d/cloud.groovy
```diff
@@ -158,120 +162,103 @@ initMap['micro-amazon'] = '''
...
         done
-        if [ -n "${DEVICE}" ]; then
-            sudo mkfs.ext2 ${DEVICE}
-            sudo mount ${DEVICE} /mnt
+        if [[ ${ARCH} == "x86_64" ]]; then
+            PKGLIST="epel-release centos-release-scl"
+        else
+            PKGLIST="epel-release"
+            until sudo yum -y update; do
+                sleep 1
```
EvgeniyPatlan, line 176: why we perform `yum update` without installing epel repo? maybe some updates would come from it?

EvgeniyPatlan (changes requested):

IaC/pxc.cd/init.groovy.d/cloud.groovy
Sudokamikaze: The same way was with PXB
Each centos-6-x32 base image is outdated, and when we install EPEL-release before any updates, it will fail to connect to the repos with SSL error

Update is needed to perform makecache after it

EvgeniyPatlan (approved):

PR #1076, opened 2021-07-02
Rename GKE clusters in cloud jobs

tplavcic:
There can be a collision if running multiple jobs on the same commit when naming the clusters.
So this will keep the PR jobs running without any addition in the name, latest jobs running on schedule with `latest` and parametrized jobs with `param` addition in the name.

### DIFF
--- a/cloud/jenkins/pxc_operator_gke_version.groovy
+++ b/cloud/jenkins/pxc_operator_gke_version.groovy
@@ -284,7 +284,7 @@ pipeline {
                 CLOUDSDK_CORE_DISABLE_PROMPTS = 1
                 GIT_SHORT_COMMIT = sh(script: 'git -C source rev-parse --short HEAD', , returnStdout: true).trim()
                 VERSION = "${env.GIT_BRANCH}-${env.GIT_SHORT_COMMIT}"
-                CLUSTER_NAME = sh(script: "echo jenkins-pxc-${GIT_SHORT_COMMIT} | tr '[:upper:]' '[:lower:]'", , returnStdout: true).trim()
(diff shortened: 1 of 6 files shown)


delgod (approved):

egegunes (approved):

cloud/jenkins/pxc_operator_gke_version.groovy
```diff
@@ -284,7 +284,7 @@ pipeline {
                 CLOUDSDK_CORE_DISABLE_PROMPTS = 1
                 GIT_SHORT_COMMIT = sh(script: 'git -C source rev-parse --short HEAD', , returnStdout: true).trim()
                 VERSION = "${env.GIT_BRANCH}-${env.GIT_SHORT_COMMIT}"
-                CLUSTER_NAME = sh(script: "echo jenkins-pxc-${GIT_SHORT_COMMIT} | tr '[:upper:]' '[:lower:]'", , returnStdout: true).trim()
+                CLUSTER_NAME = sh(script: "echo jenkins-param-pxc-${GIT_SHORT_COMMIT} | tr '[:upper:]' '[:lower:]'", , returnStdout: true).trim()
```
hors, line 287: @tplavcic please check line 378, we need to fix it not only in one place :(  Please do it for all files

hors (changes requested):

cloud/jenkins/pxc_operator_gke_version.groovy
tplavcic: @hors Thanks for catching this! I have fixed it now.

delgod (approved):

hors (approved):

PR #1097, opened 2021-07-21
 CLOUD-619: Fix missing parameters and syntax error

egegunes:


tplavcic (changes requested):
You'll need to put sergey into groovy pipeline also otherwise it will be removed after first run: https://github.com/Percona-Lab/jenkins-pipelines/blob/master/cloud/jenkins/pxc_operator_gke_latest.groovy#L151

tplavcic (approved):

PR #1121, opened 2021-07-30
PS-7828: Add job to test PS InnoDB Cluster with EC2 instances

nlaperc:
This PR moves the existing [`test-ps-innodb-cluster` job from the old Jenkins](https://jenkins.percona.com/job/test-ps-innodb-cluster/) to the ps80 Jenkins instance.

The corresponding PR for the code changes is at Percona-QA/package-testing#88; lots of additional detail there.

You can see the job running at https://ps80.cd.percona.com/job/wip-test-ps-innodb-cluster/, against the `wip-test-ps-innodb-cluster` branches.

### DIFF
--- a/ps/jenkins/test-ps-innodb-cluster.groovy
+++ b/ps/jenkins/test-ps-innodb-cluster.groovy
@@ -1,0 +1,163 @@
+library changelog: false, identifier: 'lib@master', retriever: modernSCM([
+    $class: 'GitSCMSource',
+    remote: 'https://github.com/Percona-Lab/jenkins-pipelines.git'
+]) _
+
+void installDependencies() {
+    sh '''
+        export PATH=${PATH}:~/.local/bin
+        sudo yum install -y git python3-pip jq
+        sudo amazon-linux-extras install ansible2
+        python3 -m venv venv
+        source venv/bin/activate
+        python3 -m pip install --upgrade pip setuptools wheel
+        python3 -m pip install molecule==2.22 boto boto3 paramiko testinfra
+    '''
+
+    sh '''
+        rm -rf package-testing
+        git clone https://github.com/Percona-QA/package-testing
+    '''
+}
+
+void runMoleculeAction(String action, String scenario) {
+    def awsCredentials = [
+        sshUserPrivateKey(
+            credentialsId: 'MOLECULE_AWS_PRIVATE_KEY',
+            keyFileVariable: 'MOLECULE_AWS_PRIVATE_KEY',
+            passphraseVariable: '',
+            usernameVariable: ''
+        ),
+        aws(
+            accessKeyVariable: 'AWS_ACCESS_KEY_ID',
+            credentialsId: '5d78d9c7-2188-4b16-8e31-4d5782c6ceaa',
+            secretKeyVariable: 'AWS_SECRET_ACCESS_KEY'
+        )
+    ]
+
+    withCredentials(awsCredentials) {
+        sh """
+            source venv/bin/activate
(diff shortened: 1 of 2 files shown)


ps/jenkins/test-ps-innodb-cluster.groovy
```diff
@@ -0,0 +1,163 @@
...
+]) _
+
+void installDependencies() {
+    sh '''
+        export PATH=${PATH}:~/.local/bin
+        sudo yum install -y git python3-pip jq
+        sudo amazon-linux-extras install ansible2
+        python3 -m venv venv
+        source venv/bin/activate
+        python3 -m pip install --upgrade pip setuptools wheel
```
EvgeniyPatlan, line 13: why do you update python modules via pip and not by packages from the repos?

ps/jenkins/test-ps-innodb-cluster.groovy
```diff
@@ -0,0 +1,163 @@
...
+            defaultValue: '3558242',
+            description: 'Short git hash for release'
+        )
+        choice(
+            name: 'TEST_DIST',
+            choices: [
+                'ubuntu-focal',
+                'ubuntu-bionic',
+                'ubuntu-xenial',
+                'debian-10',
```
EvgeniyPatlan, line 110: As we are currently working on adding debian 11 maybe it is high time to add into job?

ps/jenkins/test-ps-innodb-cluster.groovy
nlaperc: As you can see [right above this change](https://github.com/Percona-Lab/jenkins-pipelines/pull/1121/files/909ba9ab56efc0df1de3d119450028c4ba759006#diff-a98375b5fad48e3d1ffbd6fbe274841c5fdc16d1fe4d89a58cca4777c0990aa9R9), I am already installing the latest available version of pip (and setuptools, which is its dependency) from the repos.

However, as always, the repos are still a couple versions behind, and pip changed how it chooses which package version to install from the given requirements in October last year; [in the package installation invocation after this line](https://github.com/Percona-Lab/jenkins-pipelines/pull/1121/files/909ba9ab56efc0df1de3d119450028c4ba759006#diff-a98375b5fad48e3d1ffbd6fbe274841c5fdc16d1fe4d89a58cca4777c0990aa9R14), the upgraded version correctly picks `testinfra-3.4.0`, whereas the outdated version from the repositories picks `testinfra-6.0.0` despite it not being compatible with `molecule-2.22`.

Upgrading pip using pip is a well-established practice, and `pip` itself will display a prompt when using it, if it's out-of-date, suggesting you upgrade it in this manner. Python distributions support it. Also, [this is how pip is upgraded for other projects in this repository which install Molecule](https://github.com/Percona-Lab/jenkins-pipelines/blob/master/vars/installMolecule.groovy#L8).

ps/jenkins/test-ps-innodb-cluster.groovy
nlaperc: Sure thing, I'll work on that and let you know when it's ready.

Sudokamikaze (approved):
LGTM
But please add Debian 11 support, in case if you need an AMI ID or my assistant -- feel free to DM me on Slack

ps/jenkins/test-ps-innodb-cluster.groovy
nlaperc: The packages needed are present now and the job works as expected: https://ps80.cd.percona.com/job/wip-test-ps-innodb-cluster/98/

PR #1132, opened 2021-08-05
PXB-2572: Archive results directory for debugging purposes for both PXB

Sudokamikaze:
Builds: https://pxb.cd.percona.com/job/percona-xtrabackup-8.0-test-param-illiatests/13/

### DIFF
--- a/pxb/v2/jenkins/percona-xtrabackup-2.4-test-param.yml
+++ b/pxb/v2/jenkins/percona-xtrabackup-2.4-test-param.yml
@@ -11,9 +11,9 @@
     properties:
     - build-discarder:
         days-to-keep: -1
-        num-to-keep: 50
+        num-to-keep: 30
         artifact-days-to-keep: -1
-        artifact-num-to-keep: 50
+        artifact-num-to-keep: 30
     parameters:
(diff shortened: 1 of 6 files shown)


pxb/v2/jenkins/percona-xtrabackup-2.4-test-param.yml
```diff
@@ -11,9 +11,9 @@
     properties:
     - build-discarder:
         days-to-keep: -1
-        num-to-keep: 50
+        num-to-keep: 30
```
altmannmarcelo, line 14: Is this the number of test params to keep or the number of test pipelines?

We have 8 OS that get tested against 4 different server, resulting in 32 test pipelines for each test param

Sudokamikaze: This is only for param runs, not configurations and not of pipeline

altmannmarcelo: ack

altmannmarcelo (approved):
LGTM

PR #1151, opened 2021-08-13
ENG-1048: Grant access to PG for PSMDB

Sudokamikaze:


EvgeniyPatlan (changes requested):
we shouldn't do this as we have separate jenkins for each product and distribution tests should be also on these jenkins servers.

PR #1188, opened 2021-09-01
move pipeline from pmm-submodules repo to jenkins-pipeline

nikita-b:


atymchuk:
> To be honest it should have had name pmm2-submodules.groovy (instead of pmm-submodules.groovy) but it is more important to have the same name for job and pipeline file in my opinion
> [Percona-Lab/pmm-submodules#1996](https://github.com/Percona-Lab/pmm-submodules/pull/1996)

I would certainly agree if we didn't have a bunch of pipelines related to PMM1. But we do. If that is true, then it is much more reasonnable to name it `pmm2-submodules.groovy` and then think about how to put PMM1 and PMM2 related script apart.

atymchuk (approved):

pmm/pmm-submodules.groovy
```diff
@@ -0,0 +1,387 @@
+library changelog: false, identifier: 'lib@master', retriever: modernSCM([
```
atymchuk, line 1: Let's avoid confusion for right now and call it `pmm2-submodules.groovy`

atymchuk (changes requested):

atymchuk (approved):

PR #1193, opened 2021-09-07
fix contractor group name

nikita-b:


atymchuk (approved):
Sorry, we should have let you know in advance)

PR #1197, opened 2021-09-08
upgrade tests for a new version

nikita-b:


puneet0191 (approved):
Thanks a lot, great improvement, lets move this into a shared library too.

nikita-b:
@puneet0191 Yep, it was only a prototype :)
I moved the latest version to a library, I'm not sure about the list of versions in e2e tests and AMI upgrade...they are almost the same. Can we use the same versions for these tests?

Also, I plan to take version from pmm-submodules but it'll be in the next PR :)

pmm/pmm2-ami-upgrade-tests.groovy
```diff
@@ -60,6 +60,9 @@ void setInstanceAMIId(PMM_VERSION) {
         case "2.21.0":
             env.AMI_ID = "ami-0605d9dbdc9d6a233"
             break;
+        case "2.22.0":
+            env.AMI_ID = "ami-0c1521c259375cb43"
```
puneet0191, line 64: How are you sure about this? should we not wait until RC release day

puneet0191 (approved):
looks good to me, just one minor comment, THanks a lot for this, very useful

pmm/pmm2-ami-upgrade-tests.groovy
nikita-b: Do you mean 2.22.0 release day? We already have RC for 2.22.0 or do we add only released AMI here?

puneet0191: I think it is fine to use the RC version, I do the same, we use same for release also, but what if we need to build a new RC?

nikita-b: If we'll build a new RC we just replace ID here with a new PR :)

PR #1201, opened 2021-09-13
PMM-8625 adding additional checks on UI post client upgrade

puneet0191:


yurkovychv (approved):

pmm/pmm2-upgrade-tests.groovy
```diff
@@ -206,7 +206,7 @@ pipeline {
                     sh """
                         set -o errexit
                         set -o xtrace
-                        docker exec pmm-server yum update -y percona-release
+                        docker exec pmm-server yum update -y percona-release || true
```
nikita-b, line 209: Why do we need `true` here?

nikita-b (approved):

pmm/pmm2-upgrade-tests.groovy
puneet0191: I have to do it because we have this failing on specific versions, I could see this consistently failing on 2.16.0 with the following 
`Package percona-release-1.0-27.noarch.rpm is not signed`

PR #1236, opened 2021-10-11
PMM-8870 adding docker tag to upgrade job.

puneet0191:


pmm/pmm2-package-testing.groovy
```diff
@@ -98,7 +98,7 @@ pipeline {
             description: 'Type of Tests?',
             name: 'TESTS')
         choice(
-            choices: ['testing', 'experimental', 'main'],
+            choices: ['testing', 'experimental', 'main', 'tools_main', 'pmm2_client_main'],
```
nikita-b, line 101: I honestly don't like the name of the repos here. It seems a dash would be better: `tools-main`, `pmm2-client-main`.
But I don't have a strong opinion here;

puneet0191: okay, I think I agree with your point. Let me make changes.

PR #1237, opened 2021-10-11
PMM-7 move PMM v1 pipelines to separate directory and remove rewind for PMM 1

nikita-b:


pmm/v1/pmm-docker.groovy
```diff
@@ -12,7 +12,7 @@ pipeline {
     }
     parameters {
         string(
-            defaultValue: 'master',
+            defaultValue: 'main',
```
puneet0191, line 15: I think this should be 1.x?

PR #1267, opened 2021-10-29
PMM-9082 migrate to new farm

nikita-b:


puneet0191 (approved):

pmm/pmm2-ami.groovy
```diff
@@ -46,67 +35,42 @@ pipeline {
...
-                    sh """
-                        set -o pipefail
-                        ~/bin/packer build \
-                        -var 'pmm_client_repos=original testing' \
-                        -var 'pmm_client_repo_name=percona-testing-x86_64' \
-                        -var 'pmm2_server_repo=testing' \
-                        -only amazon-ebs -color=false packer/pmm2.json \
-                            | tee build.log
-                    """
+                sh 'make pmm2-ami-rc'
```
puneet0191, line 38: this approach could be cleaner, but for me, it decreases visibility to what repo values we are using for what parameter.

puneet0191 (approved):

pmm/pmm2-ami.groovy
nikita-b: I changed this because now we can run the build locally, just like we do in the pipeline. I think it's useful for debugging. But I agree that it slightly reduces visibility :(

atymchuk (approved):

PR #1333, opened 2021-12-13
DISTMYSQL-140: Create dev Jenkins pipeline for Orchestrator

kamil-holubicki:
https://jira.percona.com/browse/DISTMYSQL-140

Jenkins scripts for Orchestrator development purposes

### DIFF
--- a/orchestrator/jenkins/orchestrator-pipeline.groovy
+++ b/orchestrator/jenkins/orchestrator-pipeline.groovy
@@ -41,7 +41,7 @@ pipeline {
         stage('Integration Tests') {
             agent { label 'docker-32gb' }
             steps {
-                git branch: 'DISTMYSQL-140', url: 'https://github.com/kamil-holubicki/jenkins-pipelines'
+                git branch: 'master', url: 'https://github.com/Percona-Lab/jenkins-pipelines'
                 echo 'Run Orchestrator integration tests'
                 sh '''
                     # sudo is needed for better node recovery after compilation failure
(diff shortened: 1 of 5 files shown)


kamil-holubicki:
This PR is related to https://github.com/percona/orchestrator/pull/1 and the following comment applies to both.

The goal of this work is to connect Orchestrator's test environment to our Jenkins and be able to test it against various versions of Percona Server. This work should be done with minimal modification of original scripts, definitely should not be "rework from scratch".
Unfortunately, the original test environment is not flexible in the way we need, so we need to parameterize a few things (see PRs comments).
Yes, I know the current implementation is not optimal, it creates test env twice, it downloads PS tarball twice, but as said above, we just need to integrate what we have now to allow development/fixes. If necessary, we will improve it later.

The pipeline is deployed here:
https://ps80.cd.percona.com/job/mysql-orchestrator-pipeline
At this moment test failures are expected, because we need to fix DISTMYSQL-138 on top of this work.

orchestrator/jenkins/orchestrator-pipeline.groovy
```diff
@@ -0,0 +1,82 @@
...
+        skipDefaultCheckout()
+        skipStagesAfterUnstable()
+        timeout(time: 6, unit: 'DAYS')
+        buildDiscarder(logRotator(numToKeepStr: '200', artifactNumToKeepStr: '200'))
+    }
+    stages {
+        stage('Integration Tests') {
+            agent { label 'docker-32gb' }
+            steps {
+                git branch: 'DISTMYSQL-140', url: 'https://github.com/kamil-holubicki/jenkins-pipelines'
```
vorsel, line 44: Am I right that once this tested and verified, the links will point to some of our repos ?

kamil-holubicki: Yes, as it works as expected, I just committed corrected links.

EvgeniyPatlan (approved):

vorsel (approved):

PR #1352, opened 2021-12-23
PXB fails to build without libudev-devel

AgileStas:
CMake Warning at cmake/fido2.cmake:28 (MESSAGE):
  Cannot find development libraries.  You need to install the required
  packages:

    Debian/Ubuntu:              apt install libudev-dev
    RedHat/Fedora/Oracle Linux: yum install libudev-devel
    SuSE:                       zypper install libudev-devel

### DIFF
--- a/pxc/docker/install-deps
+++ b/pxc/docker/install-deps
@@ -87,12 +87,11 @@ if [ -f /usr/bin/yum ]; then
     patchelf lsof automake bzip2 gnutls gnutls-devel patch lsof socat qpress perl-XML-Simple \
     stunnel \
     cyrus-sasl-scram \
+    libudev-devel \
     "
 
     if [[ ${RHVER} -eq 8 ]]; then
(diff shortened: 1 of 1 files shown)


EvgeniyPatlan (approved):

pxc/docker/install-deps
```diff
@@ -91,6 +91,8 @@ if [ -f /usr/bin/yum ]; then
 
     if [[ ${RHVER} -eq 8 ]]; then
         PKGLIST+=" lz4 perl-Memoize perl-open valgrind-devel libubsan rpcgen libtirpc-devel boost-devel"
+	# At least PXB build fails if libudev-devel is absent
+        PKGLIST+=" libudev-devel"
```
altmannmarcelo, line 95: as per slack conversation, all distros need this library, not only centos8

altmannmarcelo (changes requested):

altmannmarcelo (approved):
LGTM

PR #1354, opened 2021-12-24
PMM-7 fix auth flow

nikita-b:
1. New versions of Github auth plugin used team slug instead of team name...
https://github.com/jenkinsci/github-oauth-plugin/commit/fcf8b7721a146e9adcc9d8bf68c88ddc5ed05e52
2. I removed `External Contractors` group because I didn't find it in GitHub and it was disabled a long time ago.


atymchuk:
> 2. I removed `External Contractors` group because I didn't find it in GitHub and it was disabled a long time ago.
No, it's still there, and we don't want to cut them off :)

atymchuk (approved):

PR #1375, opened 2022-01-17
Pmm 9461 fix pmm v1 build

nikita-b:


pmm/v1/pmm-ami.groovy
```diff
@@ -3,12 +3,12 @@ pipeline {
         specName = 'AMI'
     }
     agent {
-        label 'awscli'
+        label 'docker-farm'
```
atymchuk, line 6: I can imagine we'd call it a build farm, but too late :)

atymchuk (approved):

PR #1380, opened 2022-01-19
PMM-9393 Adding changes for custom password, and package testing playbooks

puneet0191:


pmm/pmm2-upgrade-tests.groovy
```diff
@@ -217,7 +217,7 @@ pipeline {
             }
             steps {
                 script {
-                    setupPMMClient(env.SERVER_IP, CLIENT_VERSION, 'pmm2', 'no', 'release', 'yes', 'compose_setup')
+                    setupPMMClient(env.SERVER_IP, CLIENT_VERSION, 'pmm2', 'no', 'release', 'yes', 'compose_setup', 'admin')
```
nikita-b, line 220: Can we use variables here? We can add it to `environment` section:
```
        PMM_ADMIN_PASSOWRD=admin
```

nikita-b (approved):

pmm/pmm2-upgrade-tests.groovy
puneet0191: yes, good point, will modify thank you

PR #1384, opened 2022-01-20
[ENG-1089] Remove CDN sync

AgileStas:
Repo CDN nodes were taken down and there is no need to run the clear cdn cache process anymore


AgileStas:
I did not split this PR into two (PMM and non-PMM) because it looks like we have some shared code in vars/

EvgeniyPatlan (approved):

PR #1427, opened 2022-03-01
CLOUD-687 Image CVE lookup

cap1984:


cloud/jenkins/pgo_docker_build.groovy
```diff
@@ -93,6 +93,13 @@ pipeline {
                             docker trust sign perconalab/percona-postgresql-operator:\$TAG_PREFIX-postgres-operator
                             docker trust sign perconalab/percona-postgresql-operator:\$TAG_PREFIX-pgo-deployer
                             ./e2e-tests/build
+
+                            docker run -v \$(pwd):/tmp/pgo --rm quay.io/sysdig/secure-inline-scan:2 perconalab/percona-postgresql-operator:\$TAG_PREFIX-pgo-apiserver --sysdig-token '${SYSDIG_API_KEY}' --sysdig-url https://us2.app.sysdig.com
```
hors, line 97: I think It is possible to have some file with report and please add additional stage like we have  `Check PGO Docker images`.  And we need to add it for all our images not only for pgo. Also if e.g. pgo-event has some issues we need to save the report and check next image e.g. pgo-rmdata and so on.

hors (changes requested):

hors (approved):

nmarukovich (approved):

PR #1404, opened 2022-02-02
PXC-3730 PXC8 jenkins jobs should use latest PXB 2.4/8.0 version by d…

adivinho:
Screenshot 2022-02-02 at 14 01 46

Pipelines were tested [here](https://ps3.cd.percona.com/view/PXC/job/pxc-8.0-test-param/)

PR #1444, opened 2022-03-11
PXC-3730: PXC8 jenkins jobs should use latest PXB 2.4/8.0 version by …

kamil-holubicki:
…default

https://jira.percona.com/browse/PXC-3730

It seems PR https://github.com/Percona-Lab/jenkins-pipelines/pull/1404
misses this modification. As the result PXB24 latest is not used.
Instead, the version specified in PXB24_BRANCH is used.


dutow (approved):

kamil-holubicki:
Checked on my experimental pipeline and it works fine
https://pxc.cd.percona.com/job/pxc-8.0-pipeline-parallel-mtr/

PR #1483, opened 2022-04-12
PXB-2693 - Initialize PS submodules

ldonoso:
PS 8.0.28 introduces a submodule for the first time.

If submodules are not initialized the compilation fails with:

```
CMake Error at CMakeLists.txt:1880 (ADD_SUBDIRECTORY):
  The source directory

    /mnt/jenkins/workspace/pxb80-single-platform-run/CMAKE_BUILD_TYPE/RelWithDebInfo/jenkins-pipelines/pxb/sources/pxb80/extra/libkmip

  does not contain a CMakeLists.txt file.
```

### DIFF
--- a/pxb/jenkins/checkout
+++ b/pxb/jenkins/checkout
@@ -42,6 +42,9 @@ if [ "$SOURCE_NAME" == 'PXB80' -o "$SOURCE_NAME" == 'ALL' ]; then
         if [ -n "${PXB80_REPO}" -a -n "${PXB80_BRANCH}" ]; then
             git pull origin ${PXB80_BRANCH}
         fi
+
+        git submodule deinit -f . || true
+        git submodule update --init --recursive
      popd
 fi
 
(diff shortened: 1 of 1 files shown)


satya-bodapati (approved):

pxb/jenkins/checkout
```diff
@@ -69,5 +73,9 @@ if [ "$SOURCE_NAME" == 'PXB24' -o "$SOURCE_NAME" == 'ALL' ]; then
         if [ -n "${PXB24_REPO}" -a -n "${PXB24_BRANCH}" ]; then
             git pull origin ${PXB24_BRANCH}
         fi
+
+        git submodule deinit -f . || true
+        git submodule init
+        git submodule update
```
EvgeniyPatlan, line 79: why not just `git submodule update --init`

ldonoso: @EvgeniyPatlan @satya-bodapati , I did this way because it is done so in the rest of the configurations of the pipelines. I guess the motives might be the the syntax you've proposed is not supported by new versions of git. https://stackoverflow.com/a/10168693/1449471

satya-bodapati: @EvgeniyPatlan what is the git version used in pipelines?

EvgeniyPatlan: @satya-bodapati we always install latest available in repos

satya-bodapati (approved):
LGTM!

EvgeniyPatlan (approved):

PR #1484, opened 2022-04-13
Pmm 7 add default keys for staging

nikita-b:
1. I added default SSH keys for GitHub to stagings.
2. Switched to a new docker farm for a staging job
3. Use `FAIR` price. It should reduce price on 1/3

### DIFF
--- a/vars/getSHHKeysPMM.groovy
+++ b/vars/getSHHKeysPMM.groovy
@@ -1,4 +1,5 @@
 String call() {
+    // You need to add your keys to https://github.com/settings/keys
     String sshKeys = ''
     final List additional_keys = ['nikita-b', 'nailya', 'puneet0191'] // github users
     additional_keys.each { item ->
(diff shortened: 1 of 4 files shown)


puneet0191 (approved):

vars/getSHHKeysPMM.groovy
```diff
@@ -0,0 +1,10 @@
+String call() {
+    // You need to add your keys to https://github.com/settings/keys
+    String sshKeys = ''
+    final List additional_keys = ['nikita-b', 'nailya', 'puneet0191'] // github users
```
atymchuk, line 4: Why only these people and not other? It doesn't make sense to me.

atymchuk (changes requested):

vars/getSHHKeysPMM.groovy
nikita-b: These are the people who asked. If anyone else wants, we will add :) 

I just don't want to add all people because it will slow down the build.

atymchuk (approved):

vars/getSHHKeysPMM.groovy
atymchuk: Let's see how to overcome the issue.

PR #1502, opened 2022-04-24
release-1.12.0 add env vars for pmm versions

nmarukovich:


tplavcic (approved):

tplavcic (changes requested):
Please make it parameters in the pipeline and we probably don't need both `PMM_SERVER_VER` and `IMAGE_PMM_SERVER_TAG`.

PR #1519, opened 2022-04-29
PMM-9907  fix issue with empty repo

nikita-b:
We need to check that the repository is not empty before attempting to remove packages.


EvgeniyPatlan (changes requested):
is it used for main repo too?

PR #1540, opened 2022-05-17
PMM-10034 add pipeline for LaceWork

nikita-b:


atymchuk (approved):

Oneiroi:
Good first implementation 👍 This produces a scan file with report output to `report.html`

What needs to be defined next is if this should "block" the build pipeline and if so what criteria that should occur under; for example, critical fixable issue is noted in the scan, it would make some sense for the process to cause a fail / block in the pipeline if this were the case.

PR #1572, opened 2022-06-02
Fix UI tests for PMM 2.26.0 and older

peterSirotnak:
UI tests were failing on password change on PMM 2.26.0 and older this PR fixes that
https://pmm.cd.percona.com/blue/organizations/jenkins/pmm2-ui-tests-temp/detail/pmm2-ui-tests-temp/811/pipeline

### DIFF
--- a/pmm/pmm2-ui-tests.groovy
+++ b/pmm/pmm2-ui-tests.groovy
@@ -12,6 +12,20 @@ void uploadAllureArtifacts() {
     }
 }
 
+def changeUserPasswordUtility(dockerImage) {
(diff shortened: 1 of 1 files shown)


puneet0191 (approved):
Do you have a use case of calling ui-tests on older version?

peterSirotnak:
> Do you have a use case of calling ui-tests on older version?

As part of [PMM-10097](https://jira.percona.com/browse/PMM-10097) Im automating test case that handles connection error for PMM 2.26 and older

PR #1594, opened 2022-06-15
PMM-10103 add RHEL-9 package

nikita-b:


pmm/pmm2-client-autobuild-arm.groovy
```diff
@@ -69,6 +69,7 @@ pipeline {
                     steps {
                         sh './build/bin/build-client-rpm centos:7'
                         sh './build/bin/build-client-rpm rockylinux:8'
+                        sh './build/bin/build-client-rpm almalinux:9.0'
```
LenzGr, line 72: Is there a reason why we're using a different Red Hat derivative for each major version?
Would it make sense to consolidate this to just using Oracle Linux, for example?

nikita-b: @LenzGr Oracle Linux doesn't have the final 9.0 version yet. 9.0 preview was released a couple of days ago.

Rocky also was chosen before we decided to use Oracle Linux.

We use these images only for building and I think it'll be easy to migrate to OL later.

atymchuk (approved):

pmm/pmm2-client-autobuild-arm.groovy
vorsel: @nikita-b 
+ for @LenzGr, why not to use RHEL9 AMI for builds and ubi9 for doker?

IaC/pmm.cd/init.groovy.d/cloud.groovy
```diff
@@ -317,6 +325,7 @@ devMap['docker']            = '/dev/xvda=:8:true:gp2,/dev/xvdd=:80:true:gp2'
 labelMap = [:]
 labelMap['min-centos-7-x64']  = 'min-centos-7-x64'
 labelMap['min-centos-8-x64']  = 'min-centos-8-x64'
+labelMap['min-centos-9-x64']  = 'min-centos-9-x64'
```
vorsel, line 328: @nikita-b it's quite confusing to have a "centos-9" label here, and completely different system under the hood

pmm/pmm2-client-autobuild-arm.groovy
EvgeniyPatlan: one more +1 to comments above

vars/uploadRPM.groovy
```diff
@@ -35,6 +36,11 @@ def call() {
                         `find . -name '*.el8.noarch.rpm' -o -name '*.el8.x86_64.rpm'` \
                         ${USER}@repo.ci.percona.com:\${path_to_build}/binary/redhat/8/x86_64/
                 fi
+                if [ `find . -name '*.el9.noarch.rpm' -o -name '*.el9.x86_64.rpm' | wc -l` -gt 0 ]; then
+                    scp -o StrictHostKeyChecking=no -i ${KEY_PATH} \
+                        `find . -name '*.el9.noarch.rpm' -o -name '*.el9.x86_64.rpm'` \
+                        ${USER}@repo.ci.percona.com:\${path_to_build}/binary/redhat/9/x86_64/
+                fi
```
EvgeniyPatlan, line 43: we are adding the same block the fourth time. the only difference is the os number.
So I suggest adding the loop and reusing code instead of just adding new blocks

EvgeniyPatlan (changes requested):

pmm/pmm2-client-autobuild-arm.groovy
EvgeniyPatlan: @LenzGr as @vorsel mentioned ubi9 is available

nikita-b: @vorsel @EvgeniyPatlan @LenzGr For UBI we need to change our build process. It's the same for RHEL7-9 now but for UBI we'll need to change some things. For example, we use EPEL repos and we'll need to install it in another way.

I agree that it'll be cool to use UBI everywhere but it seems a good task for a separate ticket.

Also, RHEL on AWS is not free: https://aws.amazon.com/partners/redhat/faqs/
We'll need to pay additional money to Redhut or use a developer license (it's additional bureaucracy)

We use these images for testing purposes only (we just install our package) and I think it's not worth additional resources :)

IaC/pmm.cd/init.groovy.d/cloud.groovy
nikita-b: These labels are not only for me but for QA for example.

Different labels will be less obvious for other teams. I can change it to `min-rhel-7-x64` and add notice to docs.

vars/uploadRPM.groovy
nikita-b: Good point. Fixed!

denisok (approved):

LenzGr (approved):
LGTM!

IaC/pmm.cd/init.groovy.d/cloud.groovy
EvgeniyPatlan: Well QA should also update their pipelines

pmm/pmm2-client-autobuild-arm.groovy
EvgeniyPatlan: It is not for testing! Packages will go to prod

EvgeniyPatlan (changes requested):

EvgeniyPatlan (approved):

vorsel (approved):

PR #1596, opened 2022-06-15
PSQLADM-396 proxy sql jenkins job is failing on centos 8

adivinho:
Screenshot 2022-06-15 at 18 20 41

### DIFF
--- a/proxysql/build-binary-proxysql
+++ b/proxysql/build-binary-proxysql
@@ -20,7 +20,6 @@ if [ -f /usr/bin/yum ]; then
         sudo ./bootstrap
         sudo make
         sudo make install
-        PATH=$PATH:/usr/local/bin
         sudo ln -s /usr/local/bin/cmake /usr/bin/cmake
         sudo cmake --version
     fi
--- a/proxysql/build-binary-proxysql
+++ b/proxysql/build-binary-proxysql
@@ -12,7 +12,17 @@ set -o xtrace
 if [ -f /usr/bin/yum ]; then
     RHEL=$(rpm --eval %rhel)
     if [[ $RHEL -eq 8 ]]; then
-        sudo yum -y install cmake python2 gnutls-devel libtool || true
+        sudo yum -y install python2 gnutls-devel libtool || true
+        sudo yum remove -y cmake
+        sudo wget https://github.com/Kitware/CMake/releases/download/v3.19.8/cmake-3.19.8.tar.gz
+        sudo tar -zxvf cmake-3.19.8.tar.gz
+        cd cmake-3.19.8/
+        sudo ./bootstrap
+        sudo make
+        sudo make install
+        PATH=$PATH:/usr/local/bin
+        sudo ln -s /usr/local/bin/cmake /usr/bin/cmake
+        sudo cmake --version
     fi
     sudo yum -y install wget
     if [[ $RHEL -eq 7 ]]; then
@@ -36,6 +46,7 @@ sudo wget https://dl.google.com/go/go1.17.7.linux-amd64.tar.gz
 # Hence using the default $PATH that is available to every new session (both parent and child script)
 sudo tar -C /usr/local/ -xzf go1.17.7.linux-amd64.tar.gz
 sudo cp /usr/local/go/bin/go /usr/local/bin/go
+sudo ln -s /usr/local/go/bin/go /usr/bin/go
 
 WORKDIR_ABS=$(cd ${1:-./build/proxysql}; pwd -P)
 SOURCE_DIR=$(cd ${2:-./build/proxysql}; pwd -P)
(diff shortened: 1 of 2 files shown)


proxysql/build-binary-proxysql
```diff
@@ -12,7 +12,17 @@ set -o xtrace
...
-        sudo yum -y install cmake python2 gnutls-devel libtool || true
+        sudo yum -y install python2 gnutls-devel libtool || true
+        sudo yum remove -y cmake
+        sudo wget https://github.com/Kitware/CMake/releases/download/v3.19.8/cmake-3.19.8.tar.gz
+        sudo tar -zxvf cmake-3.19.8.tar.gz
+        cd cmake-3.19.8/
+        sudo ./bootstrap
+        sudo make
+        sudo make install
+        PATH=$PATH:/usr/local/bin
```
mohitj1988, line 23: Do we need this step? IIRC /usr/local/bin is already in the PATH variable

adivinho: Yes. This line is redundant. I have removed it.

proxysql/build-binary-proxysql
```diff
@@ -12,7 +12,16 @@ set -o xtrace
...
     RHEL=$(rpm --eval %rhel)
     if [[ $RHEL -eq 8 ]]; then
-        sudo yum -y install cmake python2 gnutls-devel libtool || true
+        sudo yum -y install python2 gnutls-devel libtool || true
+        sudo yum remove -y cmake
+        sudo wget https://github.com/Kitware/CMake/releases/download/v3.19.8/cmake-3.19.8.tar.gz
+        sudo tar -zxvf cmake-3.19.8.tar.gz
+        cd cmake-3.19.8/
+        sudo ./bootstrap
+        sudo make
```
mohitj1988, line 21: Why not club both lines together
sudo make install

proxysql/build-binary-proxysql
```diff
@@ -12,7 +12,16 @@ set -o xtrace
...
+        sudo make install
+        sudo ln -s /usr/local/bin/cmake /usr/bin/cmake
```
mohitj1988, line 23: Is there a specific reason for creating this softlink ?

proxysql/build-binary-proxysql
```diff
@@ -36,6 +45,7 @@ sudo wget https://dl.google.com/go/go1.17.7.linux-amd64.tar.gz
 # Hence using the default $PATH that is available to every new session (both parent and child script)
 sudo tar -C /usr/local/ -xzf go1.17.7.linux-amd64.tar.gz
 sudo cp /usr/local/go/bin/go /usr/local/bin/go
+sudo ln -s /usr/local/go/bin/go /usr/bin/go
```
mohitj1988, line 48: Why are we creating this softlink when we already have go installed at
/usr/local/bin and this path exists in $PATH already

proxysql/build-binary-proxysql
```diff
@@ -12,7 +12,16 @@ set -o xtrace
 if [ -f /usr/bin/yum ]; then
     RHEL=$(rpm --eval %rhel)
     if [[ $RHEL -eq 8 ]]; then
-        sudo yum -y install cmake python2 gnutls-devel libtool || true
+        sudo yum -y install python2 gnutls-devel libtool || true
+        sudo yum remove -y cmake
+        sudo wget https://github.com/Kitware/CMake/releases/download/v3.19.8/cmake-3.19.8.tar.gz
+        sudo tar -zxvf cmake-3.19.8.tar.gz
```
mohitj1988, line 18: Remove the verbose option -v.

proxysql/build-binary-proxysql
adivinho: I have tested on the same docker image that is used for building.
Screenshot 2022-06-16 at 17 58 08

proxysql/build-binary-proxysql
adivinho: Screenshot 2022-06-16 at 18 01 43

proxysql/build-binary-proxysql
adivinho: Such commands sequence are taken from the official cmake site.
Screenshot 2022-06-16 at 18 08 12

proxysql/build-binary-proxysql
mohitj1988: Ah, now I understand why this behaviour is seen. You can retain the softlinks but let me explain the root cause.

Earlier, during the same shell session, cmake existed and that executable was found in /usr/bin, but that was uninstalled.

You then installed another cmake executable in /usr/local/bin.

The bash shell caches the first location that it finds for any external command that you use, which means it doesn't have to do an expensive search for the executable the next time you use the same command. The downside of this is that it won't bother to look again when you, later, install another executable by the same name, even if it's done in a directory that occurs earlier in $PATH than the original location.

The solution to this issue is to empty the cached locations of executables that bash keeps. This is done with 
(hash -r) . To only forget the location of the cmake executable, use hash -d cmake.

proxysql/build-binary-proxysql
mohitj1988: You can retain the softlinks

mohitj1988 (approved):
Looks good

PR #1681, opened 2022-07-29
PMM-7 pass pmm client version as env var

beatahandzelova:
We need to pass PERCONA_TEST_DBAAS_PMM_CLIENT=perconalab/pmm-client:dev-latest as default due to bug https://jira.percona.com/browse/PMM-10361 to make dbaas tests fail less

Before: https://pmm.cd.percona.com/blue/organizations/jenkins/pmm2-dbaas-e2e-tests/detail/pmm2-dbaas-e2e-tests/1336/tests
After: https://pmm.cd.percona.com/blue/organizations/jenkins/pmm2-dbaas-e2e-tests-beata-temp/detail/pmm2-dbaas-e2e-tests-beata-temp/4/tests

### DIFF
--- a/pmm/pmm2-dbaas-e2e-tests.groovy
+++ b/pmm/pmm2-dbaas-e2e-tests.groovy
@@ -8,7 +8,7 @@ void runStagingServer(String DOCKER_VERSION, CLIENT_VERSION, CLIENT_INSTANCE, SE
         string(name: 'DOCKER_VERSION', value: DOCKER_VERSION),
         string(name: 'CLIENT_VERSION', value: CLIENT_VERSION),
         string(name: 'CLIENT_INSTANCE', value: CLIENT_INSTANCE),
-        string(name: 'DOCKER_ENV_VARIABLE', value: '-e PMM_DEBUG=1 -e ENABLE_DBAAS=1 -e PERCONA_TEST_VERSION_SERVICE_URL=https://check-dev.percona.com/versions/v1'),
+        string(name: 'DOCKER_ENV_VARIABLE', value: DOCKER_ENV_VARIABLE),
         string(name: 'SERVER_IP', value: SERVER_IP),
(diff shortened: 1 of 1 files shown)


pmm/pmm2-dbaas-e2e-tests.groovy
```diff
@@ -65,6 +65,10 @@ pipeline {
             defaultValue: 'dev-latest',
             description: 'PMM Client version',
             name: 'CLIENT_VERSION')
+        string(
+            defaultValue: '-e PMM_DEBUG=1 -e ENABLE_DBAAS=1 -e PERCONA_TEST_VERSION_SERVICE_URL=https://check-dev.percona.com/versions/v1 -e PERCONA_TEST_DBAAS_PMM_CLIENT=perconalab/pmm-client:dev-latest',
```
puneet0191, line 69: Why hardcode this? We could use environment variable to set this value, so user could pass custom value?

puneet0191 (approved):
Just one note

pmm/pmm2-dbaas-e2e-tests.groovy
beatahandzelova: Wdym? This was hardcoded before (-e PMM_DEBUG=1 -e ENABLE_DBAAS=1 -e PERCONA_TEST_VERSION_SERVICE_URL=https://check-dev.percona.com/versions/v1) were always used when 
pmm2-dbaas-e2e-tests job was triggered. Now we actually have the option to change the env variables as you want. Just putting PERCONA_TEST_DBAAS_PMM_CLIENT=perconalab/pmm-client:dev-latest as default because it makes sense with default dev-latest server.
Screenshot from 2022-07-29 15-12-30

puneet0191: Sorry I was thinking of having a seperate client docker tag field and using that, but it is going to over complicate, please go ahead

PR #1694, opened 2022-08-04
ProxySQL compilation is failing on CentOS 7 due to below Error:

mohitj1988:
checking whether compiler supports -Wall... Can't locate IPC/Cmd.pm

Solution:
Installed the missing perl module perl-IPC-Cmd


kaushikpuneet07 (approved):
LGTM

PR #1715, opened 2022-08-15
PS-8347 add parameter for mysql shell installation for ps80

panchal-yash:
MYSQL Shell installation parameter for ps80

### DIFF
--- a/ps/jenkins/package-testing-ps-8.0.groovy
+++ b/ps/jenkins/package-testing-ps-8.0.groovy
@@ -75,7 +75,7 @@ pipeline {
         choice(
             name: "install_mysql_shell",
             choices: ["yes", "no"],
-            description: "install and check mysql shell for ps80 option"
+            description: "install and check mysql-shell for ps80"
         )
 
     }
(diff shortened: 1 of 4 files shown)


ps/jenkins/package-testing-ps-8.0.groovy
```diff
@@ -70,6 +71,13 @@ pipeline {
             choices: ["yes", "no"],
             description: "check warning in client_test"
         )
+        
+        choice(
+            name: "install_mysql_shell",
+            choices: ["yes", "no"],
+            description: "install and check mysql shell for ps80 option"
```
eleo007, line 78: Is it only for ps80? It seems we have discussed that 1 option will be used for client_check and ps80?

panchal-yash: Yes this parameter will be used by both ps80 and client_check both.

eleo007: Then we should update option name and remove " for ps80 option"

eleo007 (approved):

PR #1750, opened 2022-09-14
PMM-7 Temp Remove Junit step

puneet0191:


atymchuk (approved):
Let's merge it until we find how to fix it.

PR #1810, opened 2022-10-21
CLOUD-705 delete unused clusters

nmarukovich:


cloud/gcp-functions/cmd/clusters.go
```diff
@@ -0,0 +1,68 @@
...
+        "os"
+        "time"
+        "strconv"
+
+        "google.golang.org/api/compute/v1"
+        "google.golang.org/api/container/v1"
+)
+
+func CleanClusters(w http.ResponseWriter, r *http.Request) {
+    fmt.Fprint(w, "Hello, World test!")
```
hors, line 17: please remove all debug messages and move PR from draft

cloud/gcp-functions/cmd/clusters.go
```diff
@@ -0,0 +1,66 @@
...
+            if err != nil {
+                log.Fatal(err)
+            }
+            for _, cluster := range clustersReq.Clusters {
+                creationTime, err := time.Parse(time.RFC3339, cluster.CreateTime)
+                if err != nil {
+                    log.Fatal(err)
+                }
+
+                deleteClusterAfterHours := cluster.ResourceLabels["delete-cluster-after-hours"]
```
egegunes, line 45: you can check if label exists and continue the for loop if not. you can eliminate one level of indentation this way.

```suggestion
                deleteClusterAfterHours, ok := cluster.ResourceLabels["delete-cluster-after-hours"]
                if !ok {
                    continue
                }         
```

cloud/gcp-functions/cmd/clusters.go
```diff
@@ -0,0 +1,66 @@
...
+
+                if len(deleteClusterAfterHours) != 0 {
+                    if deleteClusterAfterHours, err := strconv.ParseInt(deleteClusterAfterHours, 10, 64); err == nil {
```
egegunes, line 48: rather than checking if error is nil, it's better to check if error is not nil and handle the error.

if there's something wrong with parsing the label value to int, we should log the error and continue. this way we can understand the value that causes the problem and also we can eliminate another level of indentation in the loop

cloud/gcp-functions/cmd/clusters.go
```diff
@@ -0,0 +1,66 @@
...
+                        if int64(math.Round(currentTime.Sub(creationTime).Hours())) > deleteClusterAfterHours {
+                            resp, err := containerService.Projects.Zones.Clusters.Delete(project, zone.Name, cluster.Name).Context(ctx).Do()
+                            if err != nil {
+                                return fmt.Errorf("delete cluster: %v", err)
```
egegunes, line 52: should we stop execution or continue to other clusters?

cloud/gcp-functions/cmd/clusters.go
```diff
@@ -0,0 +1,66 @@
...
+                            }
+                            fmt.Printf("%s\n", resp.Status)
```
egegunes, line 54: why don't we use log for these messages? we'll get timestamps and we can get rid of `\n`s at the end

egegunes (changes requested):

cloud/gcp-functions/cmd/clusters.go
```diff
@@ -0,0 +1,66 @@
...
+    if err != nil {
+        log.Fatal("Error: create service", err)
+    }
+    containerService, errContainer := container.NewService(ctx)
+    if errContainer != nil {
+        log.Fatal("Error: create containerService", errContainer)
+    }
+
+    project := os.Getenv("GCP_DEV_PROJECT")
+    zonesReq := computeService.Zones.List(project)
```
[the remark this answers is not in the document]
nmarukovich: I just try it the same as we use in disks) Just for consistency

cloud/gcp-functions/cmd/clusters.go
```diff
@@ -0,0 +1,69 @@
...
+                }
+
+                deleteClusterAfterHours, ok := cluster.ResourceLabels["delete-cluster-after-hours"]
+                if !ok {
+                    continue
+                }
+
+                deleteClusterAfterHoursInt, err := strconv.ParseInt(deleteClusterAfterHours, 10, 64)
+                if err != nil {
+                    log.Fatal(err)
```
egegunes, line 50: `log.Fatal` will stop execution, just use `log.Println` or better `log.Printf("parse label value: %v", err)`

it'll be easier which call failed if you give them proper messages

egegunes (changes requested):

cloud/gcp-functions/cmd/clusters.go
nmarukovich: ok!

egegunes (approved):

inelpandzic (approved):

PR #1824, opened 2022-10-26
PMM-7 check the validity of envvars

atymchuk:
This is to ensure clean exit for the post stage, i.e. prevent errors like the one below:
`
Error when executing always post condition:
11:56:44  groovy.lang.MissingPropertyException: No such property: BATS_TESTS_URL for class: groovy.lang.Binding
`

### DIFF
--- a/pmm/pmm2-submodules.groovy
+++ b/pmm/pmm2-submodules.groovy
@@ -339,7 +339,7 @@ pipeline {
                         slackSend channel: '#pmm-ci', color: '#00FF00', message: "[${JOB_NAME}]: build finished - ${IMAGE}"
                     }
                 } else {
-                    if(env.API_TESTS_RESULT != "SUCCESS" && API_TESTS_URL) {
+                    if(env.API_TESTS_RESULT != "SUCCESS" && env.API_TESTS_URL) {
                         addComment("API tests have failed, Please check: API: ${API_TESTS_URL}")
                     }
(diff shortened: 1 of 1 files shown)


pmm/pmm2-submodules.groovy
```diff
@@ -339,20 +339,18 @@ pipeline {
...
                     }
-                    if(env.UI_TESTS_RESULT != "SUCCESS") {
+                    if(env.UI_TESTS_RESULT != "SUCCESS" && env.UI_TESTS_URL) {
                         addComment("UI tests have failed, Please check: UI: ${UI_TESTS_URL}")
                     }
                     slackSend channel: '#pmm-ci', color: '#FF0000', message: "[${JOB_NAME}]: build ${currentBuild.result} build job link: ${BUILD_URL}"
                 }
             }
-            sh 'sudo make clean'
-            deleteDir()
```
EvgeniyPatlan, line 355: Why do you remove cleanup steps?
If it is needed please also add it to PR descripttion

atymchuk: Because such cleanup is totally redundant for an instance that will die once the job is finished. We explain it greater detail in this doc https://github.com/Percona-Lab/jenkins-pipelines/blob/master/pmm/README.md#agents

atymchuk: The `agent-amd64` and `agent-arm64` are one-off agents.

EvgeniyPatlan (approved):

PR #1855, opened 2022-11-07
PMM-10117 migrate QAN to monorepo

atymchuk:


pmm/pmm2-release-candidate.groovy
```diff
@@ -42,7 +42,6 @@ void runPMM2OVFBuild(String SUBMODULES_GIT_BRANCH, String RELEASE_CANDIDATE) {
 def pmm_submodules() {
     return [
         "pmm",
-        "qan-api2",
```
PavelKhripkov, line 45: What exactly happens when we remove this line? Don't we exclude the module from the build process?

pmm/pmm2-release.groovy
```diff
@@ -504,7 +504,6 @@ ENDSSH
                         declare -A repos=(
                             ["percona-grafana"]="percona-platform/grafana"
                             ["percona-dashboards"]="percona/grafana-dashboards"
-                            ["percona-qan-api2"]="percona/qan-api2"
```
PavelKhripkov, line 507: Not sure about what this line is responsible for. Looks like we're dropping a repo but don't provide any replacement.

pmm/pmm2-release-candidate.groovy
atymchuk: Yes, we do, but not the module, but the repo. It'll now be build from `percona/pmm` vs `percona/qan-api2`.

pmm/pmm2-release.groovy
atymchuk: It creates a tag on those repos. In this case it's already done on line 508.

PR #1887, opened 2022-11-29
PMM-7 nginx fix for upgrade

yurkovychv:


puneet0191 (approved):
Thanks a lot, how about handling for AMI and OVF?

PR #1892, opened 2022-12-01
ENG-1164 Removes support for centos:6 and use cmake3 for centos:7

VarunNagaraju:
https://jira.percona.com/browse/ENG-1164

* Disables some necessary PXC and PXB jobs from running tests on centos:6 since it is already EOL.
* Makes sure that cmake3 is being used to compile PXB in centos:7 images.

### DIFF
--- a/pxb/jenkins/build-binary-pxb24
+++ b/pxb/jenkins/build-binary-pxb24
@@ -75,6 +75,16 @@ if [[ "$COMPILER" != "default" ]]; then
     export CXX=$(echo ${COMPILER} | sed -e 's/gcc/g++/; s/clang/clang++/')
 fi
 
+# CentOS 7
+RHVER=0
+if [[ -f /etc/redhat-release ]]; then
+    RHVER="$(rpm --eval %rhel)"
+fi
+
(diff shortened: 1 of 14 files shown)


pxb/jenkins/build-binary-pxb24
```diff
@@ -75,6 +75,12 @@ if [[ "$COMPILER" != "default" ]]; then
     export CXX=$(echo ${COMPILER} | sed -e 's/gcc/g++/; s/clang/clang++/')
 fi
 
+# CentOS 7
```
kamil-holubicki, line 78: On oraclelinux:9 there was no lsb_release, I had to do
```
RHVER=0
if [[ -f /etc/redhat-release ]]; then
    RHVER="$(rpm --eval %rhel)"
fi

if [[ ${RHVER} -eq 7 ]]; then
    JOB_CMAKE='cmake3'
fi
```
I know that we are not supporting 5.7 on oraclelinux:9, but who knows if we will not be asked to... (discussion ongoing)

vorsel (approved):

PR #1916, opened 2022-12-16
ENG 1169: PXC 5.6 jenkins job improvements

VarunNagaraju:
https://jira.percona.com/browse/ENG-1169

* Updates PXC 5.6 pipeline with to use PXB 2.4 scripts instead of 2.3 scripts along with updating the default PXB version parameter.
* Removes centos-06 from the configuration matrix.

### DIFF
--- a/pxc/jenkins/pxc56-pipeline.groovy
+++ b/pxc/jenkins/pxc56-pipeline.groovy
@@ -1,4 +1,5 @@
-pipeline_timeout = 10
+def JENKINS_SCRIPTS_BRANCH = 'master'
+def JENKINS_SCRIPTS_REPO = 'https://github.com/Percona-Lab/jenkins-pipelines'
 
 pipeline {
     parameters {
@@ -18,13 +19,13 @@ pipeline {
             name: 'USE_PR') 
         string(
             defaultValue: 'https://github.com/percona/percona-xtrabackup',
-            description: 'URL to PXB23 repository',
-            name: 'PXB23_REPO',
+            description: 'URL to PXB24 repository',
+            name: 'PXB24_REPO',
             trim: true)
         string(
-            defaultValue: 'percona-xtrabackup-2.4.20',
+            defaultValue: 'percona-xtrabackup-2.4.27',
             description: 'Tag/Branch for PXC repository',
-            name: 'PXB23_BRANCH',
+            name: 'PXB24_BRANCH',
             trim: true)
 	    string(
 	        defaultValue: 'https://github.com/percona/galera',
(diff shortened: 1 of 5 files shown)


pxc/jenkins/pxc56-pipeline.groovy
```diff
@@ -256,7 +256,7 @@ pipeline {
                             sudo git -C sources reset --hard || :
                             sudo git -C sources clean -xdf   || :
 
-                            until aws s3 cp --no-progress s3://pxc-build-cache/${BUILD_TAG}/pxb23.tar.gz ./pxc/sources/pxc/results/pxb23.tar.gz; do
+                            until aws s3 cp --no-progress s3://pxc-build-cache/${BUILD_TAG}/pxb24.tar.gz ./pxc/sources/pxc/results/pxb24.tar.gz; do
```
kamil-holubicki, line 259: tgz bundle has been renamed, and I think it will cause test script execution failure
https://github.com/Percona-Lab/jenkins-pipelines/blob/master/pxc/local/test-binary-pxc56#L21

VarunNagaraju: Updated the test-binary-pxc56 script.

kamil-holubicki (commented):
Looks fine, but please test it as we spoke on Slack.
1. Create new pipeline pointing to scripts from your branch
2. install it in Jenkins using jenkins-jobs tool
3. test -> paste here links
4. once approve, delete the test pipeline from jenkins using jenkins-jobs tool
Note: Be aware to not mess up with existing jobs :)

VarunNagaraju:
https://pxc.cd.percona.com/job/varun-pxc-5.6-param/3
https://pxc.cd.percona.com/job/varun-pxc-5.6-pipeline/12/

pxc/jenkins/pxc56-pipeline.groovy
```diff
@@ -1,4 +1,6 @@
-pipeline_timeout = 10
+def pipeline_timeout = 10
```
kamil-holubicki, line 1: Just a question. I see that some of our groovy files have this `pipeline_timeout = 10` at the top, but then it is not used.
Does it make any sense? It should be then like 
```
timeout(time: pipeline_timeout, unit: 'HOURS')  {
   do something
}
```
I think it is some leftover that was copied over and over and it has no effect at all. Or do I miss something? @Percona-Lab/build-engineers

kamil-holubicki (approved):
LGTM
Please set proper JENKINS_SCRIPTS_BRANCH, JENKINS_SCRIPTS_REPO before merging.
@Percona-Lab/build-engineers are you OK with changes?

pxc/jenkins/pxc56-pipeline.groovy
vorsel: @kamil-holubicki @VarunNagaraju 
Kamil is right here, this pipeline_timeout option looks like a leftover from some experiments and could be removed.

vorsel (changes requested):

vorsel (approved):

PR #1935, opened 2023-01-05
Adding Upgrade Support for pxc package testing job

panchal-yash:
Add Amazon Linux 2


pxc/jenkins/pxc-package-testing.groovy
```diff
@@ -239,7 +260,13 @@ pipeline {
         stage("Set up") {
             steps {             
                 script{
-                    currentBuild.displayName = "${env.BUILD_NUMBER}-${params.product_to_test}-${params.node_to_test}-${params.install_repo}"                    
+                    currentBuild.displayName = "${env.BUILD_NUMBER}-${params.product_to_test}-${params.node_to_test}-${params.test_repo}-${params.test_type}"
```
eleo007, line 263: We should not fail here. We should add this logic to stages execution.

eleo007 (approved):

PR #2055, opened 2023-04-05
PMM-7 fix changed default password

vyarosh:


vyarosh:
https://pmm.cd.percona.com/blue/organizations/jenkins/pmm2-package-testing/detail/pmm2-package-testing/2231/pipeline/25 - works fine

vorsel (dismissed):

talhabinrizwan (commented):
Do you agree that it's not permissible to store any type of password in the git repository, or do you have a different viewpoint, @edux ?

vyarosh:
> Do you agree that it's not permissible to store any type of password in the git repository, or do you have a different viewpoint, @edux ?

https://github.com/Percona-Lab/jenkins-pipelines/blob/master/pmm/aws-staging-start.groovy#L45 - we cannot avoid it I suppose. At least without redesign of "aws-staging-start" job

EvgeniyPatlan (changes requested):
why we cannot add password to jenkins secrets? 
it is very insecure to store passwords(any passwords) in plain text

vyarosh:
> why we cannot add password to jenkins secrets? it is very insecure to store passwords(any passwords) in plain text
That's has been planned as workaround. Main reason is that this tests job has nothing to do with credentials at all, and this password is not a real secret but a parameter which can be typed in at the `aws-staging-start`. 

I've updated the code with proper solution: obtaining password from called job.

vyarosh:
please review.

EvgeniyPatlan (approved):

talhabinrizwan (approved):

PR #2066, opened 2023-04-11
CLOUD-781 - Add parameters for test list and suite for operators

tplavcic:
This adds TEST_SUITE and TEST_LIST parameters for operator pipelines.

Instead of TEST_SUITE having a `choice` parameter we can change it just to `string` which will be more flexibile, but so this is open for discussion.


cloud/jenkins/pxc_operator_gke_version.groovy
```diff
@@ -239,6 +251,14 @@ pipeline {
         PXC_TAG = sh(script: "if [ -n \"\${IMAGE_PXC}\" ] ; then echo ${IMAGE_PXC} | awk -F':' '{print \$2}'; else echo 'main'; fi", , returnStdout: true).trim()
     }
     parameters {
+        choice(
+            choices: ['run-release.csv', 'run-distro.csv', 'run-minikube.csv', 'run-pr.csv'],
```
hors, line 255: Why do we need to have 'run-minikube.csv' for gke job?

hors (changes requested):

cloud/jenkins/pxc_operator_gke_version.groovy
tplavcic: The intention was to put all the files which exist, nothing else. I will remove minikube from non-minikube, but should we remove the run-pr.csv as well since these jobs will not be used for PR runs?

cap1984 (dismissed):

nmarukovich (approved):

PR #2068, opened 2023-04-11
proxysql: add ol9 and jammy to groovy

eleo007:


pxc/jenkins/proxysql-package-testing.groovy
```diff
@@ -100,9 +100,10 @@ pipeline {
             choices: [
                 'min-centos-7-x64',
                 'min-ol-8-x64',
+                'min-ol-8-x64',
```
kaushikpuneet07: It should be Ol9

kaushikpuneet07 (approved):
LGTM

PR #2114, opened 2023-05-25
Add a job for creating on demand OpenShift clusters

tplavcic:


cloud/jenkins/create_openshift_cluster.groovy
```diff
@@ -0,0 +1,152 @@
...
+                kubectl get svc --all-namespaces || true
+
+                /usr/local/bin/openshift-install destroy cluster --dir=./openshift/$CLUSTER_SUFFIX || true
+            """
+        }
+    }
+}
+
+pipeline {
+    parameters {
```
hors, line 94: maybe we need to set name as well to have possibility to start two or more clusters for different people

hors (changes requested):
please create the task for it

cloud/jenkins/create_openshift_cluster.groovy
hors: @tplavcic should we have kubeconfig as an artifact?

tplavcic: name has build number so it is unique, I don't propose we go with some custom names unless you have some specific reason other than being unique

hors: ok, I see. Thsnks.

cloud/jenkins/create_openshift_cluster.groovy
```diff
@@ -0,0 +1,152 @@
...
+                        | sudo tar -C /usr/local/bin  --wildcards -zxvpf -
+                '''
+            }
+        }
+        stage('Create cluster') {
+            parallel {
+                stage('cluster1') {
+                    steps {
+                        createCluster('cluster1')
+                        sleep(time:"${params.CLUSTER_DURATION}",unit:"HOURS")
```
hors, line 137: What if the spot instance dies? Who will remove the cluster?

hors (changes requested):

cloud/jenkins/create_openshift_cluster.groovy
tplavcic: nobody, just like in our normal tests

tplavcic: as a solution we can use jenkins master for running these jobs or specify non spot instances (pricier)

hors: @tplavcic The difference is you can run it for 12 hours and spot will have more time to die. I suggest to have  https://cloud.cd.percona.com/job/pxc-operator-aws-remove-openshift  officially

PR #2150, opened 2023-06-14
PMM-7 increase sleep after docker way upgrade

yurkovychv:


atymchuk:
@yurkovychv Could you share what problem this solves? Is there an option to check the health and continue once the environment is healthy? We may need a timeout though so that we could bail out when it's taking too long. We do have examples in many places.

puneet0191:
I am going to close this for time being.

PR #2211, opened 2023-07-14
PMM-7 launch pmm-agent without sudo

atymchuk:
Problem:
In our test environment we install pmm2-client with sudo, which makes all directories and files owned by the `root` user. This is problematic, since, in real production environments, this is almost never the case, or should not be the case for security reasons.

Solution:
Leverage `PMM_DIR`, `PMM_USER`, and `PMM_GROUP` parameters so the artifacts get owned by the effective user.

Note: I'm removing as well some bloat related to PMM1, which seems to be overdue :)


puneet0191:
this will break most of the jobs because pmm-framework gets invoked with sudo as far as I recall, this will probably cause failures on all FB installations. @vyarosh  @atymchuk  so the change is not just one job :) a big huge change, for the time being, I would recommend just using temp path variable, to avoid big change we can refactor this setup later.

atymchuk:
I still don't understand why we are applying sudo instead of PMM_DIR, this is not something the customers would do in prod environments.

Also, the leftovers from the PMM1 era seem to complicate the code making it hardly readable and understandable.

I will close this one leaving it up to you guys to fix it. Replacing `sudo bash -x ./install_tarball` with `PMM_DIR="\$(pwd)/pmm2-client" bash -E -x ./install_tarball` will make things much better.

vyarosh:
I'm voting for this change! All we need to keep all running is to make sure it will work with sudo as well. And to have it we need symlincs: 
```
ln -sf ${path}/bin/pmm-admin /usr/bin/pmm-admin
ln -sf ${path}/bin/pmm-agent /usr/bin/pmm-agent
```
@puneet0191 @atymchuk Let's add symlincs and just check that staging built with tarball setup can execute pmm-admin in both ways with and without sudo. Or just let the PR hang until I'll be back and I will check it.

atymchuk:
Let's close this one in favor of https://github.com/Percona-Lab/jenkins-pipelines/pull/2213, which has been tested by this pipeline run https://pmm.cd.percona.com/job/aws-staging-start/69480/console.

atymchuk:
Let's close this one in favor of https://github.com/Percona-Lab/jenkins-pipelines/pull/2213, which has been tested by this pipeline run https://pmm.cd.percona.com/job/aws-staging-start/69480/console.

PR #2213, opened 2023-07-17
PMM-11658 move exporter configs elsewhere

atymchuk:
Ref: https://github.com/percona/pmm/pull/2341

This PR fixes the setup of PMM Client from a tarball, which does not use sudo and is installed into a directory defined by `PMM_DIR` env variable.

In our test environment we install pmm2-client with sudo, which makes all directories and files owned by the root user. This is problematic, since, in real production environments, this is almost never the case, or should not be the case for security reasons.

Solution:
Leverage PMM_DIR (if necessary, use PMM_USER and PMM_GROUP as well) parameters so the artifacts get owned by the effective user.

Note: We are removing as well the bloat related to PMM1, which seems to be overdue :)

Note: This has been tested as part of the `aws-staging-start` pipeline https://pmm.cd.percona.com/job/aws-staging-start

### DIFF
--- a/pmm/aws-staging-start.groovy
+++ b/pmm/aws-staging-start.groovy
@@ -31,11 +31,11 @@ pipeline {
     parameters {
         string(
             defaultValue: 'perconalab/pmm-server:dev-latest',
-            description: 'PMM Server docker container version (image-name:version-tag ex. perconalab/pmm-server:dev-latest or perconalab/pmm-server:pmm1-dev-latest)',
+            description: 'PMM Server docker container version (image-name:version-tag, ex: perconalab/pmm-server:dev-latest)',
             name: 'DOCKER_VERSION')
         string(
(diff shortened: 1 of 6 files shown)


pmm/aws-staging-start.groovy
```diff
@@ -3,7 +3,7 @@ import hudson.slaves.*
 import jenkins.model.Jenkins
 import hudson.plugins.sshslaves.SSHLauncher
 
-library changelog: false, identifier: 'lib@master', retriever: modernSCM([
+library changelog: false, identifier: 'lib@PMM-11658-move-exporter-configs-elsewhere', retriever: modernSCM([
```
atymchuk, line 6: TODO: revert this before merging :)

talhabinrizwan (approved):

puneet0191 (approved):

PR #2221, opened 2023-07-21
PMM-11931 Update API descriptors before RC cut.

talhabinrizwan:
PMM-11931 (Automate updating API descriptors before RC cut):

During the initial run of the RC pipeline with the 'PMM-2.0' parameter, the necessary release branches (e.g., 'pmm-2.39.0') will be created. Subsequently, the 'pmm2-rewind-submodules-fb' pipeline will be executed, with the newly created branch being passed to it. To streamline the process, we have implemented the following steps:

1. When there is an API descriptor change, and it's the first time running the RC pipeline, we will commit the changes and trigger the 'pmm2-submodule-rewind' pipeline. This pipeline will run on the 'PMM-2.0' branch, handling the updates effectively.

2. For subsequent runs of the RC pipeline, if API descriptor changes are detected, we will commit these changes directly to the respective branch (e.g., 'pmm-2.39.0'). However, there is no need to run 'pmm2-rewind-submodules-fb' again, as it's already handled in a later stage of the process.

@atymchuk - I've added the following check in newly created "Rewind Submodules" stage:
 **_env.REMOVE_RELEASE_BRANCH == 'no'_**
 
 Do we need this check in the "Update API descriptors" stage also?


atymchuk:
> I've added the following check in newly created "Rewind Submodules" stage:
env.REMOVE_RELEASE_BRANCH == 'no'
> Do we need this check in the "Update API descriptors" stage also?

I think we do, otherwise the stage will be executed during the removal of branches.

BupycHuk (approved):

atymchuk (approved):

PR #2242, opened 2023-08-08
use pip based ansible installation for jammy issues

panchal-yash:
Reverted to stable ansible version on ubuntu jammy for avoiding kmip issues


kaushikpuneet07 (approved):
LGTM

PR #2322, opened 2023-10-11
PMM-7 reduced hardcoded lists and version dependency

vyarosh:


vars/pmmVersion.groovy
```diff
@@ -52,20 +51,32 @@ def call(String type='latest') {
...
+              script: """wget -q "https://registry.hub.docker.com/v2/repositories/perconalab/pmm-client/tags?page_size=25&name=rc" -O - | jq -r .results[].name | grep '.*.*-rc\$' | sort -V | tail -n1""",
+              returnStdout: true
+      ).trim()
+      int major = rcLatest.split('\\.')[0] as Integer
+      int minor = rcLatest.split('\\.')[1] as Integer
+      return major + "." + ++minor + ".0"
+    case 'rc':
+      sh(script: "sudo yum install -y wget jq")
+      String rcLatest = sh(
+              script: """wget -q "https://registry.hub.docker.com/v2/repositories/perconalab/pmm-client/tags?page_size=25&name=rc" -O - | jq -r .results[].name | grep '.*.*-rc\$' | sort -V | tail -n1""",
```
puneet0191, line 72: why do you check it with pmm-client?

puneet0191 (changes requested):

vars/pmmVersion.groovy
vyarosh: because pmm-client image is released in pair with server but it has no [el7/el9](https://hub.docker.com/layers/percona/pmm-server/2.40.0-el7/images/sha256-cbd835477c6b1ce5c1053c47218dd6740dca8c6d87c28a70a584a74a68180dd2) suffixes which needs to be parsed off. So the regex is just simpler.

adivinho (approved):

puneet0191 (approved):

PR #2351, opened 2023-10-31
Temporary Fix for PXB-3192

altmannmarcelo:
Centos-8 is missing annobin library:

10:20:00 CMake Error at configure.cmake:337 (MESSAGE): 10:20:00 No mysys timer support detected!
10:20:00 Call Stack (most recent call first):
10:20:00 CMakeLists.txt:1621 (INCLUDE)

This is a temporary fix until the lib is installed on the base image.


pxb/v2/local/build-binary
```diff
@@ -98,6 +98,8 @@ fi
 if [[ "${DOCKER_OS}" = "centos-8" ]]; then
   if [[ ${XB_VERSION_MAJOR} -eq 8 ]] && [[ -f /opt/rh/gcc-toolset-10/enable ]]; then
     source /opt/rh/gcc-toolset-10/enable
+    # Temporary patch for PXB-3192
+    sudo yum install -y gcc-toolset-10-annobin
```
EvgeniyPatlan, line 102: maybe you should move install command before sourcing toolset?

altmannmarcelo: Fixed

EvgeniyPatlan (approved):

PR #2391, opened 2023-11-29
PMM-12729 Do not run v2 builds for v3 based branches

surbhat1595:


pmm/pmm2-submodules.groovy
```diff
@@ -410,11 +479,14 @@ pipeline {
...
             script {
                 if (currentBuild.result == null || currentBuild.result == 'SUCCESS') {
-                    if (env.CHANGE_URL) {
+                    if (env.CHANGE_URL && env.PMM_VER =~ '^2.') {
                         unstash 'IMAGE'
                         def IMAGE = sh(returnStdout: true, script: "cat results/docker/TAG").trim()
                         slackSend channel: '#pmm-ci', color: '#00FF00', message: "[${JOB_NAME}]: build finished - ${IMAGE}"
                     }
+                    else {
+                        addComment("PMM2 Submodules skipped")
```
ademidoff, line 488: It's better to say what is being built rather than what is skipped :) 
Otherwise it will keep us guessing if v3 is actually triggered.

ademidoff:
This solution will still spin up an EC2 instance per each pipeline - v2 and v3. That's not what the ticket is about.

ademidoff (changes requested):

PR #2405, opened 2023-12-05
PMM-7 get rid of svn usage

yurkovychv:


BupycHuk (approved):
can we remove installing svn too?

talhabinrizwan (approved):

PR #2550, opened 2024-03-11
PMM-7 fix submodules dispatch

yurkovychv:


ademidoff:
@yurkovychv It should be fine to close it :)

PR #2577, opened 2024-03-25
ENG-1235 Fix environment variable propagation through scripts

VarunNagaraju:
https://perconadev.atlassian.net/browse/ENG-1235

When build-binary-pxc57 script invoked build-ps/build-binary.sh the environment variables created from the jenkins job parameters were never propagated to the build-ps/build-binary.sh script since it was being invoked through a sudo session.

Adds -E to the script invocation in order to preserve the environment variables through out the scripts.

Also, updates the default value of PXB24_BRANCH jenkins job parameter.

### DIFF
--- a/pxc/local/build-binary-pxc57
+++ b/pxc/local/build-binary-pxc57
@@ -56,7 +56,6 @@ TARGET_ARCH="$(uname -m)"
 # ------------------------------------------------------------------------------
 # Set Debug options
 # ------------------------------------------------------------------------------
-export CMAKE_OPTS=""
 if [[ "${CMAKE_BUILD_TYPE}" = "Debug" ]]; then
     BUILD_COMMENT+="-debug"
     CMAKE_OPTS+=" -DDEBUG_EXTNAME=ON -DWITH_DEBUG=ON"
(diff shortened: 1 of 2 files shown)


pxc/local/build-binary-pxc57
```diff
@@ -69,6 +68,8 @@ if [[ "${WITH_ASAN}" = "yes" ]]; then
     CMAKE_OPTS+=" -DWITH_ASAN=ON"
     BUILD_COMMENT+="-asan"
 fi
+export CMAKE_OPTS=${CMAKE_OPTS}
+export SCONS_ARGS=${SCONS_ARGS}
```
kamil-holubicki, line 72: What about exporting other variables like `BUILD_COMMENT` that are used by `build-binary.sh`?

VarunNagaraju: @kamil-holubicki The other remaining variables along with BUILD_COMMENT don't come from the Jenkins job parameters and they are computed again in build-binary.sh. But, I see that "-asan" is not added to the comment properly in build-binary.sh. So, I've exported BUILD_COMMENT and the rest as well before we remove the duplications(hopefully in the future).

vorsel (approved):

PR #2581, opened 2024-03-28
ENG-1235 Adds "--sanitize" to MTR arguments for ASAN runs

VarunNagaraju:
https://perconadev.atlassian.net/browse/ENG-1235


pxc/local/test-binary-pxc57
```diff
@@ -73,6 +73,10 @@ if [[ $MTR_ARGS == *"--big-test"* ]] || [[ $MTR_ARGS == *"--only-big-test"* ]];
     TESTCASE_TIMEOUT=$((TESTCASE_TIMEOUT * 2))
 fi
 
+if [[ $WITH_ASAN == "YES" ]]; then
```
kamil-holubicki, line 76: I think we need to export WITH_ASAN in run-test57.sh as well

VarunNagaraju: Done.

PR #2626, opened 2024-05-01
Add PXC 57 EOL support

panchal-yash:
Add Support for Major and Minor Uprade
Mods for innovation lts for Minor Upgrade support
Add new jenkins node for each parallel test to ensure no env overriding


kaushikpuneet07 (approved):
LGTM

PR #2652, opened 2024-05-24
PMM-13090: replace environment variables for v3

idoqo:


pmm/v3/pmm3-api-tests.groovy
```diff
@@ -85,9 +85,8 @@ pipeline {
                 sh '''
                     docker run -d \
                     -e PMM_DEBUG=1 \
-                    -e PERCONA_TEST_CHECKS_INTERVAL=10s \
```
ademidoff, line 88: Just one question: do we want to drop this variable?

ademidoff (approved):

pmm/v3/pmm3-api-tests.groovy
idoqo: Yeah since it's not used anywhere in PMM atm

BupycHuk (approved):

PR #2708, opened 2024-07-01
K8SPG-597 add anotation to conform to the latest 1.30 EKS changes

ptankov:


cloud/jenkins/pgo_operator_eks_latest.groovy
```diff
@@ -218,6 +213,7 @@ EOF
             export KUBECONFIG=/tmp/${CLUSTER_NAME}-${CLUSTER_SUFFIX}
             export PATH=/home/ec2-user/.local/bin:$PATH
             eksctl create cluster -f cluster-${CLUSTER_SUFFIX}.yaml
+            kubectl annotate storageclass gp2 storageclass.kubernetes.io/is-default-class=true
```
hors, line 216: we need to add this annotation for all EKS jenkins jobs

cloud/jenkins/pgo_operator_eks_latest.groovy
```diff
@@ -24,13 +24,8 @@ void prepareNode() {
         echo \$(kubectl kuttl --version) is installed
 
         curl -sL https://github.com/eksctl-io/eksctl/releases/latest/download/eksctl_\$(uname -s)_amd64.tar.gz | sudo tar -C /usr/local/bin -xzf - && sudo chmod +x /usr/local/bin/eksctl
-
-        sudo yum install -y https://repo.percona.com/yum/percona-release-latest.noarch.rpm || true
-        sudo percona-release enable-only tools
     """
-}
 
-void prepareSources() {
```
hors, line 33: I can't understand why it was done :( we logically split two preparation stages not you suggest to have one big function...

hors (changes requested):

cloud/jenkins/pgo_operator_eks_latest.groovy
```diff
@@ -440,36 +440,24 @@ pipeline {
         stage('Run Tests') {
             parallel {
                 stage('cluster1') {
-                    agent {
```
hors, line 448: I am not sure that it is a  good idea because all tests will be run on the same agent in parallel and you will have a lot of conflicts

hors (changes requested):

cloud/jenkins/pgo_operator_eks_latest.groovy
ptankov: We used to have the `prepareSources()` function just for the sake of it. What do I mean? We have used to have 2 functions: `prepareNode()` and `prepareSources()` and then we just call them one by one, like this:

```
    stages {
        stage('Prepare node') {
            steps {
                prepareNode()
                prepareSources()
            }
        }
```

There is absolutely no reason for that. I just combined the 2 functions into one: `prepareNode()`.

cloud/jenkins/pgo_operator_eks_latest.groovy
ptankov: I will, but in another PR. I need to fix this pipeline first, so that I can complete the post-release steps for PG operator, which include creating a release tag in this repo `pg-operator-2.4.0`.

cloud/jenkins/pgo_operator_eks_latest.groovy
ptankov: I am experimenting to see why it is needed to have separate blocks of code like this:
```
                    agent {
                        label 'docker'
                    }
```
in every stage V.S. having a common block in the beginning, like this:
```
    agent {
        label 'docker'
    }
```
like all the rest of the pipelines for the other operators. It looks like a copy/paste error for me, but I want to be sure.

ptankov: OK, it looks like PG and PS use new node for each parallel stage - it is some limitation of kuttl. The presence of
```
                    agent {
                        label 'docker'
                    }
```
makes the new node. You are right about this, I will fix it.

hors (approved):

PR #2733, opened 2024-07-11
PMM-7 Remove EL7 support from PMM Client v3

ademidoff:


talhabinrizwan (approved):

BupycHuk (approved):
since we disabled ubuntu:focal and debian:buster for pmm3-client-autobuild-arm should we do the same for pmm3-client-autobuild?

ademidoff:
> since we disabled ubuntu:focal and debian:buster for pmm3-client-autobuild-arm should we do the same for pmm3-client-autobuild?

We did that for ARM pipeline to make its roundtrip faster while it's still in development. IDK if we can drop the same for v3 PMM Client. Can we?

PR #2793, opened 2024-08-09
PMM-7 remove latest tags from v3 release

ademidoff:
Let's remove those commands asap since we have identified them as harmful, since they could overwrite the tags of product versions that have already been released.


ademidoff (commented):
Let's remove those commands asap since we have identified them as harmful, that could overwrite the tags of product versions that have already been released.

talhabinrizwan (approved):

PR #2870, opened 2024-09-12
CLOUD-858 delete elastic ips

nmarukovich:


cloud/aws-functions/orphaned_openshift_eip.py
```diff
@@ -0,0 +1,73 @@
...
+
+    if "team" not in tags_dict.keys() or (
+        "team" in tags_dict.keys() and tags_dict["team"] != "cloud"
+    ):
+        return False
+
+    if "delete-cluster-after-hours" not in tags_dict.keys():
+        return True
+    instance_lifetime = float(tags_dict["delete-cluster-after-hours"])
+    current_time = datetime.datetime.now().timestamp()
```
jvpasinatto, line 25: Nit: Perhaps move these two lines to line 30 to be after the try catch and closer to the if , where they are used

jvpasinatto (changes requested):

cloud/aws-functions/orphaned_openshift_eip.py
nmarukovich: I can't see any problem to keep it like this. I mean it's just a couple of rows not hundreds :)

cloud/aws-functions/orphaned_openshift_eip.py
```diff
@@ -0,0 +1,73 @@
...
+        return False
+    if (current_time - creation_time) / 3600 > instance_lifetime:
+        return True
+
+    return False
+
+
+def get_ip_to_release(client, aws_region):
+    ips_for_release = []
+    ips = client.describe_addresses()["Addresses"]
```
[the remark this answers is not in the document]
nmarukovich: Not sure that we need it. If the list is empty we check it on the next step. What type of error do you think we should check?

jvpasinatto (approved):

PR #2893, opened 2024-09-24
CLOUD-855 Set currentBuild.description automatically - PG

ptankov:


jvpasinatto (approved):

cloud/jenkins/pgo_operator_aks_latest.groovy
```diff
@@ -41,9 +41,7 @@ void prepareNode() {
     if ("$PGO_POSTGRES_IMAGE") {
         currentBuild.description = "$GIT_BRANCH-$PLATFORM_VER-CW_$CLUSTER_WIDE-" + "$PGO_POSTGRES_IMAGE".split(":")[1]
     }
-}
 
-void prepareSources() {
```
eleo007, line 46: Why do we remove prepareSources step? We will execute it for 5 times in this case: on stage 'Prepare node' and on each 'clusterX' stage.

ptankov: you're right, fixed

PR #3069, opened 2024-12-17
pxb84 package testing job mods:

panchal-yash:
use ps80 instance currently.

archive instance logs for mysql (error log)

use install molecule bookworm

use ps80 jenkins creds

add pro parameter

pro/normal

add arm os and comment deb-11 and focal

add rhel-8

add skips for pxb_84 pro for deb-11, focal, rhel-8 and ol-8

add os for ms-84

mods for pxb84 pro support


kaushikpuneet07 (approved):
LGTM

adivinho (approved):

PR #3084, opened 2024-12-30
PMM-13634 Add Watchtower FB

talhabinrizwan:


pmm/v3/pmm3-submodules.groovy
```diff
@@ -176,6 +176,14 @@ pipeline {
                 archiveArtifacts 'results/docker/TAG'
             }
         }
+        stage('Build watchtower container') {
+            steps {
+                build job: 'pmm3-watchtower-autobuild', parameters: [
+                    string(name: 'GIT_BRANCH', value: params.PMM_BRANCH),
+                    string(name: 'TAG_TYPE', value: "perconalab/pmm-watchtower-fb:${params.BRANCH_NAME}-${env.FB_COMMIT}")
```
ademidoff, line 183: I believe branch names (i.e. the value of `BRANCH_NAME`) can get very long oftentimes, so the docker tag would be way too long :) I suggest we use the same approach as in Client and Server feature builds, which is to use either the PR number or the build number instead.

pmm/v3/pmm3-submodules.groovy
```diff
@@ -190,7 +198,7 @@ pipeline {
                         def STAGING_URL = "https://pmm.cd.percona.com/job/pmm3-aws-staging-start/parambuild/"
 
                         def payload = [
-                          body: "Server docker: ${IMAGE}\nClient docker: ${CLIENT_IMAGE}\nClient tarball: ${CLIENT_URL}\nStaging instance: ${STAGING_URL}?DOCKER_VERSION=${IMAGE}&CLIENT_VERSION=${CLIENT_URL}"
+                          body: "Server docker: ${IMAGE}\nClient docker: ${CLIENT_IMAGE}\nWatchtower docker: perconalab/pmm-watchtower-fb:${params.BRANCH_NAME}-${env.FB_COMMIT}\nClient tarball: ${CLIENT_URL}\nStaging instance: ${STAGING_URL}?DOCKER_VERSION=${IMAGE}&CLIENT_VERSION=${CLIENT_URL}"
```
ademidoff, line 201: Same here.

pmm/v3/pmm3-submodules.groovy
talhabinrizwan: I have aligned it with the server and client Docker images.

ademidoff: Hm... I can't see any docker images with tags containing the branch names at https://hub.docker.com/r/perconalab/pmm-server/tags. It seems like it's not the case, at least not for the Server docker images.

ademidoff (approved):

PR #3109, opened 2025-01-16
CLOUD-875 [PS] Combine xxx-latest and an xxx-version pipelines to use common Groovy code

ptankov:


cloud/jenkins/job-pso-eks.yml
eleo007, line 1: What does `job` mean in the name? why we do not use the same naming as we have for .groovy?

ptankov: Because it makes for a better sorting in the file browser or the IDE, e.g. in VSCode. When you search for the file (Ctrl + P) and start typing `pso-...` you get only the groovy script. Whereas, if the pipeline definitions don't star with `job...` you get mixed results. It's purely fo
Screenshot_2025-01-23_22-18-27
Screenshot_2025-01-23_22-18-43
r comfortable usage when you want to distinguish files.

eleo007: For me it is easier to search when the names of `groovy` and `yml` are the same (like we had, the difference was in `-` vs `_`) because it is easy to know what name of `yml` file to search for specific `groovy` file.

eleo007: I did ....

nmarukovich (approved):

eleo007 (approved):

PR #3187, opened 2025-02-27
PML-66 Adding Jenkinsfile for mlink testing

keithquinnpercona:


mlink/mlink-functional-tests.groovy
```diff
@@ -0,0 +1,107 @@
...
+                                                branch: params.MLINK_BRANCH,
+                                                poll: false
+                                    }
+                                }
+
+                                sh """
+                                    cd psmdb-testing/mlink
+                                    docker-compose build
+                                    docker-compose up -d
+                                    docker-compose run test pytest test_basic_sync_rs.py -v -s --junitxml=junit.xml -k ${TEST} || true
```
sandraromanchenko, line 74: Why do check test_basic_sync_rs.py only? When test are extended, there would be much more of those.
+
-k ${TEST} is unnecessary as mentioned above there is no need to divide on test type for now

mlink/mlink-functional-tests.groovy
```diff
@@ -0,0 +1,107 @@
...
+        }
+        stage ('Run tests') {
+            matrix {
+                agent {
+                    label "${params.instance}"
+                }
+                axes {
+                    axis {
+                        name 'TEST'
+                        values 'replicaset', 'shard'
```
sandraromanchenko, line 36: For the nearest future mlink would support RS setup only, so there is no need for now to divide runs based on setup type. I think it's more practical to run the same set of tests for different PSMDB versions, like here https://github.com/Percona-Lab/jenkins-pipelines/blob/master/pbm/pbm-functional-tests-full.groovy#L37-L40

sandraromanchenko (changes requested):

sandraromanchenko (approved):

sandraromanchenko (approved):

olexandr-havryliak (approved):

EvgeniyPatlan (approved):

PR #3242, opened 2025-03-28
pdpxc upgrade parallel pxco intergation trigger addition

panchal-yash:
ready pdpxc upgrade parallel job for pr

### DIFF
--- a/pdpxc/pdpxc-pxco-integration-scheduler.groovy
+++ b/pdpxc/pdpxc-pxco-integration-scheduler.groovy
@@ -1,0 +1,106 @@
+pipeline {
+    agent {
+        label 'docker'
+    }
+    triggers {
+        cron('H * * * *') 
+    }
+    environment {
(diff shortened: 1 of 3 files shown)


pdpxc/pdpxc-pxco-integration-scheduler.groovy
```diff
@@ -0,0 +1,106 @@
...
+                                    env."${key}" = value  // Workaround to avoid putAt()
+                                }
+                            }
+                            echo "Environment variables set!"
+                            
+                            build(
+                                job: 'pxco-gke-1',
+                                parameters: [
+                                    string(name: "TEST_SUITE", value: "run-distro.csv"),
+                                    string(name: "IGNORE_PREVIOUS_RUN", value: "YES"),
```
eleo007, line 81: Please pass NO here. This will be the first run with those images and branch so all the test should run and rerunning failed ones will be easier.
Can we implement retry if the 1st run fails?

PR #3253, opened 2025-04-10
PMM-13922 Publish rpmbuild to Dockerhub

ademidoff:
https://perconadev.atlassian.net/browse/PMM-13922


talhabinrizwan (approved):

BupycHuk (commented):
Code looks good, but would be good to hear feedback from @EvgeniyPatlan on if it's ok to make it public

BupycHuk (approved):

ademidoff:
> Code looks good, but would be good to hear feedback from @EvgeniyPatlan on if it's ok to make it public

FYI: it has been public since day one.

PR #3304, opened 2025-04-30
PMM-13758 Migrate to dnf

talhabinrizwan:


ademidoff (commented):
Looks good, but I still have two questions:

1. Why not fix the test pipelines in the same PR? It's a simple find/replace anyway and it will allow us to close the ticket.
2. Should we change the path [here](https://github.com/Percona-Lab/jenkins-pipelines/blob/6c0de45ae549abe1c4a9098669aee12638fbbc8a/pmm/v3/pmm3-release.groovy#L51)

BupycHuk (approved):

talhabinrizwan:
> Looks good, but I still have two questions:
> 
> 1. Why not fix the test pipelines in the same PR? It's a simple find/replace anyway and it will allow us to close the ticket.
I've also updated the test pipelines.

> 2. Should we change the path [here](https://github.com/Percona-Lab/jenkins-pipelines/blob/6c0de45ae549abe1c4a9098669aee12638fbbc8a/pmm/v3/pmm3-release.groovy#L51)?
We aren't going to change that as its an actual repo path. Also, we don't change the repo paths like /etc/yum.repo.d

PR #3356, opened 2025-05-28
PKG-473: Add ccache shared library functions for build optimization

nogueiraanderson:
This PR adds shared library functions for managing ccache in Jenkins builds, enabling 80-90% compilation time reduction across Percona products.

## Changes
- Add `ccacheDownload()` function for downloading cache from S3
- Add `ccacheUpload()` function for uploading cache to S3

## Benefits
- Reusable across all Percona product pipelines
- Configurable with sensible defaults
- Automatic project name extraction and cache key generation
- Docker-based cache statistics display

## Related
- JIRA: [PKG-473](https://perconadev.atlassian.net/browse/PKG-473)
- ps-build PR: https://github.com/Percona-Lab/ps-build/pull/463

## Usage Example
```groovy
// Download cache before build
ccacheDownload()

// Run build
build("./docker/run-build")

// Upload cache after build
ccacheUpload()
```

## Configuration
Both functions accept optional configuration map:
```groovy
ccacheDownload([
    awsCredentialsId: 'custom-creds',
    s3Bucket: 's3://custom-bucket/',
    cacheSize: '5G'
])
```


EvgeniyPatlan (approved):

inikep:
`ccache` output depends on compiler, build type (Debug/RelWithDebInfo) and server version (or branch name)
In Azure Pipelines we use different caches/buckets for different builds so caches will not be polluted and we will get more hits:  
https://github.com/percona/percona-server/blob/8.0/azure-pipelines.yml#L384-L390
It's described at https://learn.microsoft.com/en-us/azure/devops/pipelines/tasks/reference/cache-v2?view=azure-pipelines
There is even `restoreKeys` param there if primary key misses.
Can we do something similar?

inikep (approved):
LGTM

PR #3362, opened 2025-05-30
PMM-7 fix default query source for nightly

yurkovychv:


peterSirotnak (changes requested):
You also need to add this extra pxc service into step: 'Check agent status'

peterSirotnak (approved):

PR #3385, opened 2025-06-13
CLOUD-910: Add docker image to run tests on pipelines

valmiranogueira:
Add github action to build the docker image every monday. Latest version is always updated and patch version is incremented.

### DIFF
--- a/pmm/v3/pmm3-aws-staging-start.groovy
+++ b/pmm/v3/pmm3-aws-staging-start.groovy
@@ -3,11 +3,6 @@ import hudson.slaves.*
 import jenkins.model.Jenkins
 import hudson.plugins.sshslaves.SSHLauncher
 
-library changelog: false, identifier: 'lib@master', retriever: modernSCM([
-    $class: 'GitSCMSource',
-    remote: 'https://github.com/Percona-Lab/jenkins-pipelines.git'
-]) _
-
 library changelog: false, identifier: 'v3lib@master', retriever: modernSCM(
   scm: [$class: 'GitSCMSource', remote: 'https://github.com/Percona-Lab/jenkins-pipelines.git'],
   libraryPath: 'pmm/v3/'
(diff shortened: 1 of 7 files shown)


nogueiraanderson (changes requested):
## PR Review - Request Changes

Thank you for implementing the base Docker image for cloud testing pipelines. The concept is excellent and will help improve pipeline efficiency. However, I've identified several critical issues that need to be addressed before merging.

### Critical Issues

1. **Architecture Mismatch (BLOCKER)**
   - The Dockerfile hardcodes `ARCH="amd64"` while the workflow builds for both `linux/amd64` and `linux/arm64`
   - This will cause ARM64 builds to download incorrect binaries and fail
   - **Fix**: Use Docker BuildKit's `TARGETARCH` variable instead of hardcoded architecture

2. **Security Vulnerabilities**
   - Container runs as root user (security risk)
   - Multiple `curl | bash` commands without signature verification
   - Most tools use "latest" versions (supply chain risk)

3. **Platform-Specific Issues**
   - Google Cloud SDK download hardcoded to `x86_64` (line 58)
   - kubectl-assert installation uses `|| true` which hides errors

### 📋 Recommendations

**Immediate fixes required:**
```dockerfile
# Replace hardcoded ARCH with BuildKit variable
ARG TARGETARCH
ENV ARCH=${TARGETARCH}

# Add non-root user
RUN useradd -m -s /bin/bash clouduser
USER clouduser

# Pin versions and verify checksums
ARG KUBECTL_VERSION=v1.31.3
RUN curl -sLO "https://dl.k8s.io/release/${KUBECTL_VERSION}/bin/linux/${ARCH}/kubectl" && \
    curl -sLO "https://dl.k8s.io/release/${KUBECTL_VERSION}/bin/linux/${ARCH}/kubectl.sha256" && \
    echo "$(cat kubectl.sha256) kubectl" | sha256sum --check
```

valmiranogueira:
@nogueiraanderson, I could fix some of the previous comments. Regarding the latest versions: it’s the same approach for pipelines—the idea is to always keep the newest versions in updated images. For example, the GitHub Action calculates a new tag for every build.

Related to the user: I set it to jenkins because Jenkins is the default user for the containers, but it can be changed if needed.

Signature verification could be implemented; however, it may not be strictly necessary, since corrupted .tgz files cannot be extracted anyway.

valmiranogueira:
Added checksum verification.

PR #3491, opened 2025-08-19
PMM-7 Refactor launchSpotInstance function

ademidoff:


ademidoff:
> @ademidoff The `DAYS` variable is used in `runSpotInstance.groovy` at line 117 but it's not passed as a parameter to the function.
> 
> ```groovy
> Key=stop-after-days,Value=${DAYS} ```
> 
> This variable appears to be expected from the calling context, but it would be better to pass it as an explicit parameter to the `runSpotInstance` function for clarity and to avoid potential undefined variable errors.
> ```

We have a number of `global` variables, that are necessary to run such pipelines - `VM_NAME`, `DAYS`, and `OWNER`. They are used in many places and therefore fixing the function signatures would be out of scope of this ticket. 

I do agree that explicitly passing the variables is much easier to reason about and support, but that would be require us to make changes to almost every other pipeline and then test them all.

PR #3512, opened 2025-08-28
OpenShift version checker utility script

nogueiraanderson:
## Summary

This PR creates an OpenShift version checker utility script.

### DIFF
--- a/scripts/openshift-patch-versions-checker.sh
+++ b/scripts/openshift-patch-versions-checker.sh
@@ -44,8 +44,8 @@ echo ""
 VERSIONS=("4.19" "4.18" "4.17" "4.16")
 CHANNELS=("stable" "fast" "candidate")
 
-# Function to get last 5 patches for a version/channel
-get_last_5_patches() {
+# Function to fetch and cache API response
+fetch_api_response() {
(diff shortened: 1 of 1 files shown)


EvgeniyPatlan (changes requested):

nogueiraanderson:
@EvgeniyPatlan Thanks for pointing out the duplicate API calls and race condition issue. I've optimized the script to address this concern.

**Changes:**
- Refactored to make a single API call per channel/version combination
- Cache the API response and parse it for both count and version list
- This eliminates the race condition where count and versions could be inconsistent

**Performance Improvements:**

| Metric | Old Version | New Version | Improvement |
|--------|------------|-------------|-------------|
| Execution Time | 35.7 seconds | 24.9 seconds | 30% faster |
| API Calls | 36 calls | 24 calls | 33% reduction |
| API Calls per Version | 3 (2 in main loop + 1 in summary) | 2 (1 in main loop + 1 in summary) | 33% reduction |

**Testing:**
- Tested on macOS: Working correctly
- Tested on Alpine Linux (Docker): Working correctly  
- Output remains identical to the previous version

PR #3545, opened 2025-09-15
pxc pt active choice plugin implementation:

panchal-yash:
* dynamic os selection based on product selection
* remove al-2023 for non pro pxc
* rhel-10, rhel-10-arm addition
* pxc_innovation_lts remove unsupported os


kaushikpuneet07 (approved):
LGTM

PR #3573, opened 2025-10-02
PXC-4760 Add support for running PXC 5.7 jenkins jobs on Hetzner

VarunNagaraju:
https://perconadev.atlassian.net/browse/PXC-4760

* Adds new CLOUD parameter to PXC 5.7 to run tests on hetzner.
* Replaces WITH_ASAN parameter with ANALYZER_OPTS with more options to compile with either ASAN or valgrind.
* Adds missing valgrind specific options to be passed with MTR.
* Adds Ubuntu noble as an additional platform to run tests on.
* Disables compilation with -DWITH_MAN_PAGES=OFF since it is not necessary for PXC testing and sphinx is not used for docs anymore which caused PXB compilation failure in Ubuntu noble.
* Minor refactoring.


pxc/jenkins/pxc57-pipeline.groovy
```diff
@@ -51,10 +59,17 @@ pipeline {
...
+           choices: [
+             '',
+             '-DWITH_ASAN=ON -DWITH_ASAN_SCOPE=ON',
+             '-DWITH_ASAN=ON',
+             '-DWITH_ASAN=ON -DWITH_ASAN_SCOPE=ON -DWITH_UBSAN=ON',
+             '-DWITH_ASAN=ON -DWITH_UBSAN=ON',
+             '-DWITH_UBSAN=ON',
+             '-DWITH_VALGRIND=ON'
+           ].join('\n'),
+           description: 'Analyzer options passed to CMake (ASAN/UBSAN/Valgrind)')
```
inikep, line 72: We should remove whole `parameters` section because it's duplicated with `pxc/jenkins/pxc57-pipeline.yml`.
In 8.0+ we removed `parameters` section many years ago and all works fine.

VarunNagaraju: Agree. 
If we're removing duplication, where should we leave the parameters section?
Leaving them in the yml file means that, the job needs to be updated(with jenkins-jobs tool) every time there is a change in any of the parameters. But, if we they're present in the groovy file, there is no need of updating the job, it will automatically picked up by jenkins.

VarunNagaraju:
@kamil-holubicki Yes, it was tested using https://pxc.cd.percona.com/job/pxc-5.7-pipeline-hetzner/ job with all the possible scenarios.

PR #3898, opened 2026-03-23
Install dependencies needed for PXC 9.6 build.

kamil-holubicki:


nogueiraanderson (changes requested):
@kamil-holubicki two findings from reviewing this PR:

**googletest URL mismatch (blocking)**

The URL downloads a `.tar.gz` but the local filename says `.zip`. Validated on CentOS Stream 9:

```
$ file googletest-release-1.17.0.zip
googletest-release-1.17.0.zip: gzip compressed data, from Unix

$ unzip -t googletest-release-1.17.0.zip
End-of-central-directory signature not found.
```

Also, the `build-binary-pxc` fallback constructs the URL `/archive/release-1.17.0.zip`, which returns HTTP 404 (googletest switched from `release-X.Y.Z` to `vX.Y.Z` tags at 1.13.0). On a cache miss the `until wget` loop hangs forever.

Suggested fix:
```bash
wget_loop 'googletest-release-1.17.0.zip' 'https://github.com/google/googletest/archive/refs/tags/v1.17.0.zip'
```

---

**Missing sanitizer packages (suggestion)**

`gcc-toolset-14-libasan-devel` and `gcc-toolset-14-libubsan-devel` are available in CentOS Stream 9 AppStream but not included here. Without them, ASAN/UBSAN sanitizer builds won't work with the 9.6 toolchain. devtoolset-15 includes them.

```
$ yum list available gcc-toolset-14-libasan-devel gcc-toolset-14-libubsan-devel
gcc-toolset-14-libasan-devel.aarch64    14.2.1-13.el9    appstream
gcc-toolset-14-libubsan-devel.aarch64   14.2.1-13.el9    appstream
```

hulyav:
Based on the discussion during the team meeting, not needed anymore.

PR #3924, opened 2026-03-27
PMM-14778 Add USE_ONDEMAND parameter to switch from spot to on-demand instances

talhabinrizwan:


ademidoff (approved):

pmm/v3/pmm3-image-scanning.groovy
```diff
@@ -4,7 +4,7 @@ This job helps run an image scan with Trivy
 
 pipeline {
     agent {
-        label 'agent-amd64-ol9'
+        label params.USE_ONDEMAND ? 'agent-amd64-ol9-ondemand' : 'agent-amd64-ol9'
```
ademidoff, line 7: IMO, we don't need this one to be on-demand :)

ademidoff:
Where can I find the code for the on-demand agent? Is it in the monorepo?

PR #3965, opened 2026-04-17
Fix eppel install for rhel10

eleo007:


pt/jenkins/pt-package-testing.groovy
```diff
@@ -19,6 +19,14 @@ setup_rhel_package_tests = { ->
     '''
 }
 
+setup_rhel10_package_tests = { ->
+    sh '''
+        sudo yum install -y https://dl.fedoraproject.org/pub/epel/epel-release-latest-10.noarch.rpm
+        sudo yum -y update
```
EvgeniyPatlan, line 25: do we really need to update everything in docker?

PR #3967, opened 2026-04-20
CLOUD-938 Add base work for image certification

valmiranogueira:
This PR adds standardized image certification pipelines for PSO, PSMDBO, PGO, and PXCO.

## What changed

- Added shared certification logic in `cloud/common/imageCertification.groovy`
- Added/updated certification pipelines for PSO, PSMDBO, PGO, and PXCO
- Created `certify_images.py` to:
  - install Go/preflight locally
  - support `amd64`, `arm64`, and `multiplatform`
  - store preflight artifacts in `preflight-results`
  - generate Jenkins JUnit XML results
  - mask sensitive tokens in logs
- Slack is only sent when at least one image passed or failed

## Certified Image Tag Pattern

Images are published to:

```text
quay.io/redhat-isv-containers/<project-id>:<tag>
```

Tag format:

```text
<release>
<release>-<component>
<release>-<component>-<source-image-version>

Examples:

1.2.0
1.2.0-pmm3
1.2.0-ps-8.4.3
1.2.0-psmdb-8.0.4-1
1.2.0-postgres-16.8
1.2.0-backup-2.0.0
```


cloud/jenkins/pso_image_certification.groovy
```diff
@@ -0,0 +1,244 @@
...
+    return images
+}
+
+def getTag(image) {
+    def parts = image.tokenize(":")
+    return parts.size() > 1 ? parts[-1] : "latest"
+}
+
+def buildTargetImage(key, image, params) {
+    def PS_OPERATOR_PROJECT_ID = "68f123a8583dd663019ea44d"
```
eleo007: I think these should be stored in jenkins creds and not shared openly in our git.

cloud/jenkins/pso_image_certification.groovy
```diff
@@ -0,0 +1,244 @@
...
+}
+
+pipeline {
+    agent {
+        label 'docker'
+    }
+
+    parameters {
+        string(name: 'BRANCH', defaultValue: '', description: 'Optional override. Defaults to release-{RELEASE}')
+        string(name: 'RELEASE', defaultValue: '1.0.0')
```
eleo007: Lets keep default as an empty string and set validation if nothing is provided - to exit with error. Otherwise by mistake we can re-appload incorrect images imho.

cloud/jenkins/pso_image_certification.groovy
```diff
@@ -0,0 +1,244 @@
...
+            defaultValue: false,
+            description: 'Skip preflight failure if image is already published'
+        )
+
+        password(
+            name: 'PYXIS_TOKEN', 
+            description: 'Access https://connect.redhat.com/account/api-keys to create one'
+        )
+
+        password(name: 'REGISTRY_USER_OPERATOR')
```
eleo007: Why not to use jenkins creds to store keys instead of entering them manually?

cloud/jenkins/pso_image_certification.groovy
```diff
@@ -0,0 +1,244 @@
...
+                            error("Image not found in release_versions: ${params.IMAGE}")
+                        }
+
+                        echo "Processing ${params.IMAGE} -> ${selectedImage}"
+
+                        def target = buildTargetImage(params.IMAGE, selectedImage, params)
+                        certifyImage(target, params)
+                    }
+                }
+            }
```
eleo007, line 173: I believe that we need post action that will:
- create artifacts with all the successfully certified images and failed ones if any.
- not sure but think would be nice to have: send message to slack channel re certif status the same as we have with tests.

cloud/jenkins/pso_image_certification.groovy
valmiranogueira: I thought it was okay to have it here since we have some id like: https://github.com/redhat-openshift-ecosystem/certified-operators/blob/main/operators/percona-server-mysql-operator-certified/ci.yaml

cloud/jenkins/pso_image_certification.groovy
valmiranogueira: Cause they need to change every release. Maybe not registry ones cause they have fixed values.

cloud/jenkins/pso_image_certification.groovy
valmiranogueira: I was thinking about not spam the slack channel, but I can add too

cloud/scripts/certify_images.py
```diff
@@ -0,0 +1,270 @@
...
+
+
+def ensure_go():
+    go_path = shutil.which("go")
+    if go_path:
+        return go_path
+
+    log("go not found, installing")
+
+    version = "1.22.5"
```
jvpasinatto: can we move this to the beginning of file and update this version?

cloud/jenkins/pso_image_certification.groovy
```diff
@@ -0,0 +1,244 @@
...
+              --dest_image ${target.dest} \
+              --component ${target.component} \
+              --platform ${platform}
+        """
+    }
+}
+
+pipeline {
+    agent {
+        label 'docker'
```
jvpasinatto: can we use hetzner  agents?

cloud/scripts/certify_images.py
```diff
@@ -0,0 +1,270 @@
+import argparse
+import subprocess
+import json
+import os
+import sys
+import shutil
+from datetime import datetime
+
```
jvpasinatto, line 9: since we will have a lot of python in this folder lets use ruff formatter to keep the same pattern as the others

cloud/scripts/certify_images.py
```diff
@@ -0,0 +1,270 @@
...
+    ], cwd=repo, env=env, quiet=True)
+
+    os.chmod("preflight", 0o755)
+    shutil.rmtree(repo)
+
+    log("preflight ready")
+
+
+def is_already_published(output: str) -> bool:
+    return "published image can't be updated" in output.lower()
```
jvpasinatto: i don't get it this func

cloud/jenkins/pso_image_certification.groovy
```diff
@@ -0,0 +1,244 @@
+def prepareSources(params) {
```
jvpasinatto: pls extract this and other common functions that can be re-used in others operators certification pipelines into a separated file. You then can load this file by something like`def helper = load "helper.groovy"`

cloud/jenkins/pso_image_certification.groovy
```diff
@@ -0,0 +1,244 @@
...
+    echo "Using branch: ${branch}"
+
+    sh """
+        set -e
+        rm -rf source
+        git clone -b ${branch} https://github.com/percona/percona-server-mysql-operator.git source
+    """
+}
+
+def loadReleaseVersions() {
```
jvpasinatto: does this also handles the platform lines?

cloud/scripts/certify_images.py
```diff
@@ -0,0 +1,270 @@
...
+
+
+def cleanup_image(image):
+    subprocess.run(["docker", "rmi", "-f", image],
+                   stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL)
+    subprocess.run(["docker", "image", "prune", "-f"],
+                   stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL)
+
+
+def prepare_image(image, dest, platform):
```
jvpasinatto, line 117: we need to make sure to not push images with already existing tags cause they will be overritten in registry

jvpasinatto (changes requested):

cloud/jenkins/pso_image_certification.groovy
valmiranogueira: Done

cloud/scripts/certify_images.py
valmiranogueira: It detects in the logic if the image has been published by the log and this can be used to skip already published items in order to certify non published images (skip failure).

cloud/scripts/certify_images.py
valmiranogueira: It can be overwritten in quay.io but not in redhat (preflight won't allow overwrite). I think it's okay to republish an image that didn't pass certification for example.

cloud/jenkins/pso_image_certification.groovy
valmiranogueira: They are skipped in the certification cause the considered items are part of image switch case, when it doesn't match there is:

        default:
            echo "Skipping ${key}"
            return null

cloud/jenkins/pso_image_certification.groovy
valmiranogueira: On it

cloud/jenkins/pso_image_certification.groovy
valmiranogueira: I think they take too long to be scheduled, I will add it as a param as other pipelines

cloud/jenkins/pso_image_certification.groovy
valmiranogueira: They don't need to change every release really, thanks!
Tokens will be part of credentials now

cloud/scripts/certify_images.py
valmiranogueira: Done

cloud/scripts/certify_images.py
valmiranogueira: I am fetching latest now

cloud/jenkins/pso_image_certification.groovy
valmiranogueira: WDYT? Still necessary to save in secrets?

eleo007 (approved):

cloud/jenkins/psmdbo_image_certification.groovy
```diff
@@ -0,0 +1,173 @@
...
+                    echo "Platform: ${params.PLATFORM}"
+                    echo "Selection: ${params.IMAGE}"
+
+                    def failedImages = []
+                    def skippedImages = []
+
+                    if (params.IMAGE == 'ALL') {
+
+                        echo "Running certification for ALL images"
+
```
jvpasinatto: here we use a diff approach from others with imagesToCertify.each is this needed? if not, pls align with others

cloud/jenkins/pso_image_certification.groovy
```diff
@@ -0,0 +1,178 @@
...
+
+        choice(name: 'JENKINS_AGENT', choices: ['Hetzner', 'AWS'], description: 'Cloud infra for build')
+    }
+
+    stages {
+        stage('Prepare Sources') {
+            steps {
+                script {
+                    certification = load "cloud/common/imageCertification.groovy"
+
```
jvpasinatto, line 104: missing `if (params.RELEASE?.trim()) {
                        currentBuild.displayName = params.RELEASE.trim()
                    }`

cloud/jenkins/pgo_image_certification.groovy
```diff
@@ -0,0 +1,194 @@
...
+        case 'IMAGE_PGBOUNCER18':
+            return target(image, containersProjectId, "${params.RELEASE}-pgbouncer-${imageTag(image)}", containersCredentials)
+
+        case 'IMAGE_BACKREST14':
+        case 'IMAGE_BACKREST15':
+        case 'IMAGE_BACKREST16':
+        case 'IMAGE_BACKREST17':
+        case 'IMAGE_BACKREST18':
+            return target(image, containersProjectId, "${params.RELEASE}-pgbackrest-${imageTag(image)}", containersCredentials)
+
```
jvpasinatto, line 60: pls add upgrade image

jvpasinatto (changes requested):

cloud/jenkins/pso_image_certification.groovy
valmiranogueira: done

cloud/jenkins/pgo_image_certification.groovy
valmiranogueira: done

cloud/jenkins/psmdbo_image_certification.groovy
valmiranogueira: done

jvpasinatto (approved):

PR #3988, opened 2026-04-23
remove -v3 from the tags of custom images

Manika-Percona:
remove -v3 from custom images as we have complete image now 
and we can move forward with just image build number

### DIFF
--- a/ppg/ppg-docker-arm-custom.groovy
+++ b/ppg/ppg-docker-arm-custom.groovy
@@ -26,7 +26,7 @@ pipeline {
         stage('Set build name'){
             steps {
                 script {
-                    currentBuild.displayName = "${params.PPG_REPO}-${env.IMAGE_VER} v3"
+                    currentBuild.displayName = "${params.PPG_REPO}-${env.IMAGE_VER}"
                 }
             }
         }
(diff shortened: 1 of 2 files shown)


surbhat1595 (approved):

rjd15372:
The PR title and description say that these changes update the custom images to include the new version of patroni, but I only see changes on the groovy files that remove the `-v3` suffix.

I missing how these changes will make the new patroni be included in the images?

Manika-Percona:
> The PR title and description say that these changes update the custom images to include the new version of patroni, but I only see changes on the groovy files that remove the `-v3` suffix.
> 
> I missing how these changes will make the new patroni be included in the images?

Jenkins groovy script does not contain the list or versions of the component which are included in the images. I mentioned about patroni in the description and title so that we could know that why we are rebuilding these images this time and that too without -v3.

rjd15372:
I see your point, but he inclusion of patroni 4.1.2 will only happen when we trigger the rebuild of images in jenkins, and not because of merging the changes in this PR.

I suggest renaming the PR title and remove the first bullet of the description because if someone else looks at this PR will get confused as I was.

Manika-Percona:
> I see your point, but he inclusion of patroni 4.1.2 will only happen when we trigger the rebuild of images in jenkins, and not because of merging the changes in this PR.
> 
> I suggest renaming the PR title and remove the first bullet of the description because if someone else looks at this PR will get confused as I was.

updated!

rjd15372 (approved):
lgtm

PR #3995, opened 2026-04-26
fix(fleet): increase EC2 worker idle termination from 3 to 15 minutes

nogueiraanderson:
## Summary

- Increases `idleTerminationMinutes` from `3` to `15` for all EC2 worker templates on pxb.cd
- Prevents premature worker termination during parallel test pipeline runs where workers appear briefly idle between pipeline stages
- Already deployed to the live instance on 2026-04-24 after an incident where 59 builds were aborted in a single run

## Context

The EC2 plugin's `ComputerRetentionWork` timer terminates workers idle longer than `idleTerminationMinutes`. With a 3-minute window, workers running parallel PXB test pipelines were killed between stages (brief idle gap while the next stage schedules), causing `AgentOfflineException` and mass build aborts.

15 minutes provides sufficient grace for stage transitions. Worst-case cost: ~$0.05/worker for 12 extra idle minutes on m5n.2xlarge spot.

### DIFF
--- a/IaC/cloud.cd/init.groovy.d/cloud.groovy
+++ b/IaC/cloud.cd/init.groovy.d/cloud.groovy
@@ -359,7 +359,7 @@ SlaveTemplate getTemplate(String OSType, String AZ) {
             new EC2Tag('Name', 'jenkins-cloud-' + OSType),
             new EC2Tag('iit-billing-tag', 'jenkins-cloud-worker')
         ],                                          // List<EC2Tag> tags
-        '3',                                        // String idleTerminationMinutes
+        '15',                                       // String idleTerminationMinutes
         0,                                          // Init minimumNumberOfInstances
         0,                                          // minimumNumberOfSpareInstances
         capMap[typeMap[OSType]],                    // String instanceCapStr
--- a/IaC/pg.cd/init.groovy.d/cloud.groovy
+++ b/IaC/pg.cd/init.groovy.d/cloud.groovy
@@ -179,7 +179,7 @@ SlaveTemplate getTemplate(String OSType, String AZ) {
             new EC2Tag('Name', 'jenkins-pg-' + OSType),
             new EC2Tag('iit-billing-tag', 'jenkins-pg-worker')
         ],                                          // List<EC2Tag> tags
-        '3',                                        // String idleTerminationMinutes
+        '15',                                       // String idleTerminationMinutes
         0,                                          // Init minimumNumberOfInstances
         0,                                          // minimumNumberOfSpareInstances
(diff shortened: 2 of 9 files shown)


nogueiraanderson:
@evgeniy You're right about the mechanism. `idleTerminationMinutes` runs AFTER a task finishes and the agent goes idle, and yes, while it's idle the same VM can pick up another task. But two things:

*1. `3` is not actually guaranteeing clean-VM-per-job today.* Only `pxc.cd` (per-label via `maxUseMap['singleUse']=1`) and `pmm.cd` (globally via `maxTotalUses=1`) actually destroy the VM after each job. The other 8 masters (`pg`, `ps3`, `ps57`, `ps80`, `psmdb`, `pxb`, `rel`, `cloud`) all pass `maxTotalUses=-1` to every `SlaveTemplate`, so the same dirty VM CAN pick up another job within whatever idle window we set. `3` just narrowed the pollution window on those 8; it never closed it. (Side note: `pmm.cd` already runs at `idleTerminationMinutes='10'` and is unaffected by this PR.)

*2. The proper EC2 plugin parameter for clean-VM-per-job is `maxTotalUses=1`*, added in ec2-plugin v1.43 (May 2019, JENKINS-8618). When set, the agent calls `setAcceptingTasks(false)` from `taskAccepted` once the counter drains, and `taskCompleted` then invokes `slaveNode.terminate()` directly via `postJobAction`, bypassing `idleTerminationMinutes` entirely. That's how `pxc.cd` does it. ([EC2RetentionStrategy.java#L413-L464](https://github.com/jenkinsci/ec2-plugin/blob/master/src/main/java/hudson/plugins/ec2/EC2RetentionStrategy.java#L413-L464), [help-maxTotalUses.html](https://github.com/jenkinsci/ec2-plugin/blob/master/src/main/resources/hudson/plugins/ec2/SlaveTemplate/help-maxTotalUses.html))

Also worth noting: Hetzner workers (the bulk of our capacity) have *no single-use mode at all* in the dNation plugin, only `IdlePeriodPolicy` and `BeforeHourWrapsPolicy`. They run with `idleMinutes=10` fleet-wide. 60% of our pipelines call `deleteDir()` for workspace cleanup (321/531 Jenkinsfiles).

If clean-VM-per-job is the real requirement for those 8 masters, the right fix is porting `pxc.cd`'s `maxUseMap` pattern to them, not keeping `idleTerminationMinutes=3` as a fragile proxy. Happy to file a follow-up ticket if you want to push that direction. The PR 3995 change is purely about scheduling efficiency for parallel/matrix builds where executors briefly idle between stages.

vorsel (approved):

PR #3997, opened 2026-04-27
PKG-1312: pre-stage real LE cert before OpenResty start (ps3 canary)

nogueiraanderson:
Tracking in [PKG-1312](https://perconadev.atlassian.net/browse/PKG-1312).

## Issue

During instance bootstrap, OpenResty starts with a self-signed cert before the real Let's Encrypt cert is restored. Clients hitting the box during that window see TLS warnings.

Affects all Jenkins masters that share this bootstrap pattern.

## Fix

Restore the real LE cert from the on-disk backup before OpenResty starts, so the trusted cert is in place from the first request. Existing renewal flow is unchanged; first-time bootstrap (no backup present) is unaffected.

[PKG-1312]: https://perconadev.atlassian.net/browse/PKG-1312?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/IaC/ps3.cd/JenkinsStack.yml
+++ b/IaC/ps3.cd/JenkinsStack.yml
@@ -797,6 +797,22 @@ Resources:
                   systemctl restart certbot-renew.timer
               }
 
+              restore_real_cert_from_backup() {
+                  # Must run before setup_nginx: OpenResty would otherwise start
+                  # with the stale self-signed cert from create_fake_ssl_cert and
+                  # serve TLS warnings until setup_letsencrypt swaps the symlinks.
+                  [[ -d /mnt/ssl_backup ]] || return 0
+                  rsync -aHSv --delete /mnt/ssl_backup/ /etc/letsencrypt/
+                  local cert=/etc/letsencrypt/live/$JENKINS_HOST/fullchain.pem
+                  local key=/etc/letsencrypt/live/$JENKINS_HOST/privkey.pem
+                  [[ -f $cert && -f $key ]] || return 0
+                  # Skip if backup cert is expired; setup_letsencrypt will provision fresh.
+                  openssl x509 -in $cert -checkend 0 -noout || return 0
+                  mkdir -p /etc/nginx/ssl
+                  ln -f -s $cert /etc/nginx/ssl/certificate.crt
+                  ln -f -s $key  /etc/nginx/ssl/certificate.key
+              }
+
               setup_dhparam() {
                   if [ ! -f /mnt/$JENKINS_HOST/ssl/dhparam-4096.pem ]; then
                       openssl dhparam -out /mnt/$JENKINS_HOST/ssl/dhparam-4096.pem 4096
@@ -856,6 +872,7 @@ Resources:
                   install_software
                   start_jenkins
                   create_fake_ssl_cert
+                  restore_real_cert_from_backup
                   setup_nginx
                   setup_dhparam
                   setup_letsencrypt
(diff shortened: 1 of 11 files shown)


EvgeniyPatlan (changes requested):
Could we fold this into create_fake_ssl_cert — use the backup if it’s valid, otherwise self-signed? That keeps main() unchanged and lets us validate before the rsync so an expired backup doesn’t pollute /etc/letsencrypt/.

nogueiraanderson:
@EvgeniyPatlan good catch on the validation order. Pushed 7dad6853: now we check `openssl x509 -checkend 0` against the backup at `/mnt/ssl_backup/live/$JENKINS_HOST/fullchain.pem` *before* the rsync, so an expired backup never reaches `/etc/letsencrypt/`.

I kept it as a separate function rather than folding into `create_fake_ssl_cert` since the operations differ (cp self-signed vs. symlink real LE cert into `/etc/letsencrypt/live/`) and the names communicate intent. Happy to merge them if you feel strongly.

IaC/ps3.cd/JenkinsStack.yml
```diff
@@ -797,6 +797,23 @@ Resources:
...
               }
 
+              restore_real_cert_from_backup() {
+                  # Must run before setup_nginx: OpenResty would otherwise start
+                  # with the stale self-signed cert from create_fake_ssl_cert and
+                  # serve TLS warnings until setup_letsencrypt swaps the symlinks.
+                  [[ -d /mnt/ssl_backup ]] || return 0
+                  local src_cert=/mnt/ssl_backup/live/$JENKINS_HOST/fullchain.pem
+                  local src_key=/mnt/ssl_backup/live/$JENKINS_HOST/privkey.pem
+                  [[ -f $src_cert && -f $src_key ]] || return 0
```
EvgeniyPatlan, line 807: better to check with -r - we can verify that the file is readable

IaC/ps3.cd/JenkinsStack.yml
```diff
@@ -797,6 +797,23 @@ Resources:
...
+                  # Validate before rsync so an expired backup does not pollute
+                  # /etc/letsencrypt/; setup_letsencrypt will provision fresh.
+                  openssl x509 -in $src_cert -checkend 0 -noout || return 0
+                  rsync -aHSv --delete /mnt/ssl_backup/ /etc/letsencrypt/
+                  mkdir -p /etc/nginx/ssl
+                  ln -f -s /etc/letsencrypt/live/$JENKINS_HOST/fullchain.pem /etc/nginx/ssl/certificate.crt
```
EvgeniyPatlan, line 813:
```suggestion
                  ln -fsn /etc/letsencrypt/live/$JENKINS_HOST/fullchain.pem /etc/nginx/ssl/certificate.crt
```

IaC/ps3.cd/JenkinsStack.yml
```diff
@@ -797,6 +797,23 @@ Resources:
...
+                  ln -f -s /etc/letsencrypt/live/$JENKINS_HOST/privkey.pem  /etc/nginx/ssl/certificate.key
```
EvgeniyPatlan, line 814:
```suggestion
                  ln -fsn /etc/letsencrypt/live/$JENKINS_HOST/privkey.pem  /etc/nginx/ssl/certificate.key
```

EvgeniyPatlan (approved):

PR #90, opened 2019-04-17
PMM-3864 Adding tests for PMM2 OVF images

puneet0191:
1. Ensures PMM2 OVF images are installed and imported via virtualbox

2. Will be used to test setup of `pmm-admin` and `pmm-agent` and run some sanity tests later.

PR #4020, opened 2026-05-04
PKG-1325: disable IPv6 early in Hetzner worker userData

nogueiraanderson:
## Changes

In `initMap['deb-docker']` userData, insert at the top (right after `set -o xtrace`):

```bash
sudo sysctl -w net.ipv6.conf.all.disable_ipv6=1 || true
sudo sysctl -w net.ipv6.conf.default.disable_ipv6=1 || true
sudo sysctl -w net.ipv6.conf.eth0.disable_ipv6=1 || true
echo "precedence ::ffff:0:0/96 100" | sudo tee -a /etc/gai.conf
```

The `gai.conf` line forces `getaddrinfo()` to prefer IPv4 — covers Go binaries that bypass kernel IPv6 state.

Applied to 9 Debian-Hetzner IaC files: `pg`, `pmm`, `ps3`, `ps57`, `ps80`, `psmdb`, `pxb`, `pxc`, `rel`. Cloud (Fedora) excluded.

Tracks [PKG-1325](https://perconadev.atlassian.net/browse/PKG-1325). Validated on `pg.cd` build [#92](https://pg.cd.percona.com/job/docker-cve-scan/92/) (UNSTABLE 7m28s, CVEs found, no network errors — vs #90 FAILURE 18m20s with `network is unreachable`).

[PKG-1325]: https://perconadev.atlassian.net/browse/PKG-1325?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/IaC/pmm.cd/init.groovy.d/htz.cloud.groovy
+++ b/IaC/pmm.cd/init.groovy.d/htz.cloud.groovy
@@ -78,6 +78,11 @@ networkMap['percona-vpc-eu'] = '10442325' // percona-vpc-eu
 initMap = [:]
 initMap['deb-docker'] = '''#!/bin/bash -x
     set -o xtrace
+    # Force IPv4 early (Hetzner -> Cloudflare CDN IPv6 routing is intermittently broken, see PKG-1325)
+    sudo sysctl -w net.ipv6.conf.all.disable_ipv6=1 || true
+    sudo sysctl -w net.ipv6.conf.default.disable_ipv6=1 || true
+    sudo sysctl -w net.ipv6.conf.eth0.disable_ipv6=1 || true
+    echo "precedence ::ffff:0:0/96 100" | sudo tee -a /etc/gai.conf
     echo -e "nameserver 9.9.9.9\nnameserver 1.1.1.1" | sudo tee /etc/resolv.conf
     echo '10.30.6.9 repo.ci.percona.com' | sudo tee -a /etc/hosts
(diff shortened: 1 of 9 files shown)


IaC/pmm.cd/init.groovy.d/htz.cloud.groovy
```diff
@@ -78,6 +78,11 @@ networkMap['percona-vpc-eu'] = '10442325' // percona-vpc-eu
 initMap = [:]
 initMap['deb-docker'] = '''#!/bin/bash -x
     set -o xtrace
+    # Force IPv4 early (Hetzner -> Cloudflare CDN IPv6 routing is intermittently broken, see PKG-1325)
+    sudo sysctl -w net.ipv6.conf.all.disable_ipv6=1 || true
+    sudo sysctl -w net.ipv6.conf.default.disable_ipv6=1 || true
+    sudo sysctl -w net.ipv6.conf.eth0.disable_ipv6=1 || true
+    echo "precedence ::ffff:0:0/96 100" | sudo tee -a /etc/gai.conf
```
EvgeniyPatlan, line 85: I am not sure if adding || true to the end really helps. Let's imagine case when we start node and all these commands failed but because of true - everything is good. So in result we will get the node that will work with ipv6.

EvgeniyPatlan (approved):

PR #4021, opened 2026-05-04
PG-2302 PG-2325 New pipelines for PPG docker release verification workflows.

Naeem-Akhter:
* PG-2302 Added a pipeline for CVE scanning of ppg docker images, official and custom.
* PG-2325 Added a pipeline for ppg official docker image verification.

### DIFF
--- a/ppg/docker-cve-scan.groovy
+++ b/ppg/docker-cve-scan.groovy
@@ -35,20 +35,7 @@ Build Report: ${env.BUILD_URL}"""
     slackSend color: color, message: buildSummary, channel: '#postgresql-test'
 }
 
-def installTrivy(String version) {
-    sh """
-        TRIVY_CHECKSUM="8b4376d5d6befe5c24d503f10ff136d9e0c49f9127a4279fd110b727929a5aa9"
-        if command -v dnf >/dev/null 2>&1; then
-            sudo dnf install -y wget tar
-        else
-            sudo apt-get -o DPkg::Lock::Timeout=300 install -y wget tar
-        fi
-        wget -q https://github.com/aquasecurity/trivy/releases/download/v${version}/trivy_${version}_Linux-64bit.tar.gz
-        echo "\${TRIVY_CHECKSUM}  trivy_${version}_Linux-64bit.tar.gz" | sha256sum -c -
-        sudo tar zxvf trivy_${version}_Linux-64bit.tar.gz -C /usr/local/bin/
-        rm -f trivy_${version}_Linux-64bit.tar.gz
-    """
-
+def writeHelperScripts() {
     writeFile file: 'scan_to_txt.py', text: '''
 import json, sys
 
@@ -316,7 +303,6 @@ docker.io/perconalab/percona-distribution-postgresql-upgrade-custom:18.3-17.9-16
 
     environment {
         PATH = "/usr/local/bin:/usr/bin:/usr/local/sbin:/usr/sbin:${env.HOME}/.local/bin"
-        TRIVY_VERSION = '0.70.0'
     }
 
     triggers {
@@ -354,7 +340,8 @@ docker.io/perconalab/percona-distribution-postgresql-upgrade-custom:18.3-17.9-16
             steps {
                 cleanWs()
                 script {
-                    installTrivy(env.TRIVY_VERSION)
+                    installTrivy(method: 'auto')
+                    writeHelperScripts()
                 }
             }
         }
--- a/ppg/docker-cve-scan.groovy
+++ b/ppg/docker-cve-scan.groovy
@@ -1,0 +1,416 @@
+library changelog: false, identifier: "lib@master", retriever: modernSCM([
+    $class: 'GitSCMSource',
+    remote: 'https://github.com/Percona-Lab/jenkins-pipelines.git'
+])
+
+def sendSlackNotification(Map groupResults) {
+    def overallStatus = currentBuild.result ?: 'SUCCESS'
+    def hasCvesAny = groupResults.values().any { it }
+    def color = (overallStatus == 'ABORTED') ? 'warning'
+              : (overallStatus == 'FAILURE')  ? 'danger'
+              : hasCvesAny                    ? 'warning'
+              :                                 'good'
+
+    def lines = []
+    groupResults.each { group, hasCves ->
+        def icon = hasCves ? ':warning:' : ':white_check_mark:'
+        def reportLink = "<${env.BUILD_URL}artifact/trivy-${group}.html|CVE Report>"
+        lines << "${icon} *${group}*: ${hasCves ? 'CVEs found' : 'Clean'} | ${reportLink}"
+    }
+
+    if (overallStatus == 'ABORTED') {
+        lines << ':no_entry: Build was cancelled.'
+    } else if (overallStatus == 'FAILURE') {
+        lines << (groupResults ? ':x: Build failed — some scans may not have completed.' : ':x: Build failed before any scans completed.')
+    }
+
+    if (!lines) {
+        lines << ':white_check_mark: All scans completed successfully.'
+    }
+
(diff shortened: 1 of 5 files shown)


ppg/docker-cve-scan.groovy
```diff
@@ -0,0 +1,432 @@
...
+            sudo dnf install -y wget tar
+        else
+            sudo apt-get -o DPkg::Lock::Timeout=300 install -y wget tar
+        fi
+        wget -q https://github.com/aquasecurity/trivy/releases/download/v${version}/trivy_${version}_Linux-64bit.tar.gz
+        echo "\${TRIVY_CHECKSUM}  trivy_${version}_Linux-64bit.tar.gz" | sha256sum -c -
+        sudo tar zxvf trivy_${version}_Linux-64bit.tar.gz -C /usr/local/bin/
+        rm -f trivy_${version}_Linux-64bit.tar.gz
+    """
+
```
EvgeniyPatlan: we have a shared library to install trivy

EvgeniyPatlan: we have a shared library to install trivy. it is created in order to keep it in one place for easier maintenance

ppg/docker-cve-scan.groovy
```diff
@@ -0,0 +1,432 @@
...
+    writeFile file: 'scan_to_txt.py', text: '''
```
EvgeniyPatlan, line 39: Trivy already produces tabular output. --format table gives you essentially this for free. Even if you want custom columns, use a Trivy template (--format template --template "@...") instead of writeFile-ing a Python script every build.

ppg/docker-cve-scan.groovy
```diff
@@ -0,0 +1,432 @@
...
+            v.get("InstalledVersion", "").ljust(col_w[3]),
+            (v.get("FixedVersion") or "n/a").ljust(col_w[4]),
+        ])
+        print(f"{row}  {v.get(\'Title\', \'\')}")
+
+if not vulns_found:
+    print("No HIGH/CRITICAL vulnerabilities found.")
+'''
+
+    writeFile file: 'scan_to_xml.py', text: '''
```
EvgeniyPatlan, line 75: The contrib junit.tpl ships with the Trivy package and is the solution: trivy image --format template --template "@contrib/junit.tpl" -o file.xml.

ppg/docker-cve-scan.groovy
```diff
@@ -0,0 +1,432 @@
...
+        failures += 1
+
+ts.set("tests", str(failures))
+ts.set("failures", str(failures))
+with open(output_file, "w") as f:
+    f.write(\'<?xml version="1.0" encoding="UTF-8"?>\\n\')
+    f.write(ET.tostring(root, encoding="unicode"))
+'''
+
+    writeFile file: 'merge_xml.py', text: '''
```
EvgeniyPatlan: Jenkins junit testResults: 'trivy-${groupName}-*.xml' accepts globs and merges automatically.

ppg/docker-cve-scan.groovy
```diff
@@ -0,0 +1,432 @@
...
+for f in sorted(f for f in glob.glob(f"trivy-{group}-*.xml") if not f.endswith("-all.xml")):
+    try:
+        for ts in ET.parse(f).getroot():
+            root.append(ts)
+    except Exception:
+        pass
+open(output_file, "w").write(ET.tostring(root, encoding="unicode"))
+'''
+
+    writeFile file: 'group_to_html.py', text: '''
```
EvgeniyPatlan, line 111: --format template --template "@contrib/html.tpl" produces an HTML report.

EvgeniyPatlan (changes requested):

ppg/docker-cve-scan.groovy
surbhat1595: @Naeem-Akhter You can use https://github.com/Percona-Lab/jenkins-pipelines/blob/hetzner/vars/installTrivy.groovy to install trivy.

ppg/docker-server-parallel-upgrade.groovy
```diff
@@ -0,0 +1,117 @@
...
+                        'rhel-10',
+                        'debian-12',
+                        'debian-13',
+                        'ubuntu-jammy',
+                        'rocky-9-arm64',
+                        'rhel-10-arm64',
+                        'debian-12-arm64',
+                        'debian-13-arm64',
+                        'ubuntu-jammy-arm64',
+                    ]
```
surbhat1595, line 87: It's not necessary to test the docker image on multiple distros since it's independent of the host OS. We can test it only on different architecture (amd64 and arm64) hosts.

ppg/docker-cve-scan.groovy
Naeem-Akhter: Good point, will switch to the shared library. Can you confirm the hetzner branch version handles custom timeout and checksum verification?

ppg/docker-cve-scan.groovy
Naeem-Akhter: We intentionally generate txt from JSON to avoid running Trivy multiple times per image. Using --format table would require a separate run. Happy to revert if multiple runs are acceptable. Why waste extra CPU cycles in running reports multiple times? Have one JSON format, convert it to XML, HTML (for easy viewing), and text files (that don't show the giggerish characters in the browser), user experience!

ppg/docker-cve-scan.groovy
Naeem-Akhter: Same as above — we derive XML from JSON to additional Trivy run per image. 
Have one JSON format, convert it to XML, consolidated HTML (for easy viewing in browser), and text files (that don't show the gibberish characters in the browser), better user experience!

ppg/docker-cve-scan.groovy
Naeem-Akhter: The contrib template generates one HTML per image/run. Our Python script produces a single consolidated report per group, covering all images and platforms, with a summary table, making it more useful for reviewing multiple images at once.

ppg/docker-cve-scan.groovy
Naeem-Akhter: You're right, this can be simplified. Will remove the merge script and use a glob pattern directly in the junit step

ppg/docker-cve-scan.groovy
Naeem-Akhter: Same as above —  I was having the timeout issue. How to handle that in the installtrivy library?

ppg/docker-server-parallel-upgrade.groovy
Naeem-Akhter: The Docker image itself is OS-independent, but the upgrade workflow being tested involves host-level interactions through Molecule (volume mounts, storage drivers etc) that can behave differently across distros. We also want to validate against actual customer environments that run on all these platforms. So we need to have provisioning to test on multiple OS.

ppg/docker-cve-scan.groovy
Naeem-Akhter: The changes have been applied as suggested.

ppg/docker-cve-scan.groovy
Naeem-Akhter: We run one Trivy scan to generate a JSON file, then use that same JSON to create XML, TXT, and HTML files, as needed. It saves time by avoiding multiple trivy runs.

ppg/docker-cve-scan.groovy
Naeem-Akhter: Now using installTrivy.groovy.  For making it available on the master branch, I also added the same file to this PR.

Naeem-Akhter: Now using installTrivy.groovy. For making it available on the master branch, I also added the same file to this PR.

surbhat1595 (approved):

EvgeniyPatlan (approved):

PR #4039, opened 2026-05-11
PMM-15022: Stabilize PMM HA ROSA pipeline

talhabinrizwan:
- Add explicit readiness gates for worker nodes, Kyverno TLS, and all PMM HA components
  - Kyverno: scale admission controller to 3 replicas + PDB, poll for TLS cert instead of blind sleep 30
  - Grant node-exporter and nonroot-v2 SCCs to monitoring service accounts
  - Tag cluster with creation-time and delete-cluster-after-hours for cost tracking and avoid node deletion
  - Fix cleanup: capture cluster ID before deletion and sweep leftover VPC endpoints, security groups, and subnet tags
  - Switch workers from 7×m5.xlarge to 3×m7i.2xlarge
  - Double pipeline timeout to 120m

### DIFF
--- a/pmm/v3/pmm3-ha-rosa.groovy
+++ b/pmm/v3/pmm3-ha-rosa.groovy
@@ -4,6 +4,14 @@ def cleanupCluster() {
         sh '''
             rosa login --token="${ROSA_TOKEN}"
 
+            # Capture cluster ID before deletion (cluster may already be partly gone if create failed midway)
+            CLUSTER_ID=""
+            if [ -f ${HOME}/cluster-id.txt ]; then
+                CLUSTER_ID=$(cat ${HOME}/cluster-id.txt)
+            elif rosa describe cluster --cluster="${CLUSTER_NAME}" --region="${REGION}" >/dev/null 2>&1; then
+                CLUSTER_ID=$(rosa describe cluster --cluster="${CLUSTER_NAME}" --region="${REGION}" -o json | jq -r '.id // empty')
+            fi
+
             if rosa describe cluster --cluster="${CLUSTER_NAME}" --region="${REGION}" >/dev/null 2>&1; then
                 rosa delete cluster --cluster="${CLUSTER_NAME}" --region="${REGION}" --yes --watch
             fi
@@ -20,6 +28,29 @@ def cleanupCluster() {
                 rosa delete oidc-provider --oidc-config-id "${OIDC_ID}" --region "${REGION}" --mode auto --yes || true
                 rosa delete oidc-config --oidc-config-id "${OIDC_ID}" --region "${REGION}" --mode auto --yes || true
             fi
+
+            # Sweep AWS resources tied to this cluster's ID (catches what ROSA may leave behind)
+            if [ -n "$CLUSTER_ID" ]; then
+                VPC_ID=$(aws ec2 describe-vpcs --region "${REGION}" \
+                    --filters "Name=tag:Name,Values=${VPC_NAME}" "Name=state,Values=available" \
+                    --query 'Vpcs[0].VpcId' --output text 2>/dev/null)
+                if [ -n "$VPC_ID" ] && [ "$VPC_ID" != "None" ]; then
+                    echo "Sweeping AWS resources for cluster ID $CLUSTER_ID"
+                    for sub in $(aws ec2 describe-subnets --region "${REGION}" --filters "Name=vpc-id,Values=$VPC_ID" --query 'Subnets[*].SubnetId' --output text | tr '\\t' '\\n'); do
+                        aws ec2 delete-tags --region "${REGION}" --resources "$sub" --tags Key="kubernetes.io/cluster/${CLUSTER_ID}" 2>/dev/null || true
+                    done
+                    for vpce in $(aws ec2 describe-vpc-endpoints --region "${REGION}" --filters "Name=vpc-id,Values=$VPC_ID" "Name=tag:kubernetes.io/cluster/${CLUSTER_ID},Values=owned" --query 'VpcEndpoints[].VpcEndpointId' --output text | tr '\\t' '\\n'); do
+                        [ -z "$vpce" ] && continue
+                        echo "  Deleting VPC endpoint $vpce"
+                        aws ec2 delete-vpc-endpoints --region "${REGION}" --vpc-endpoint-ids "$vpce" || true
+                    done
+                    for sg in $(aws ec2 describe-security-groups --region "${REGION}" --filters "Name=vpc-id,Values=$VPC_ID" "Name=group-name,Values=${CLUSTER_ID}-*" --query 'SecurityGroups[].GroupId' --output text | tr '\\t' '\\n'); do
+                        [ -z "$sg" ] && continue
+                        echo "  Deleting SG $sg"
+                        aws ec2 delete-security-group --region "${REGION}" --group-id "$sg" || true
+                    done
+                fi
+            fi
         '''
     }
 }
@@ -31,7 +62,7 @@ pipeline {
 
     options {
         buildDiscarder(logRotator(numToKeepStr: '30'))
-        timeout(time: 60, unit: 'MINUTES')
+        timeout(time: 120, unit: 'MINUTES')
         disableConcurrentBuilds()
     }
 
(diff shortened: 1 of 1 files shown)


pmm/v3/pmm3-ha-rosa.groovy
```diff
@@ -540,7 +604,20 @@ EOF
...
+                            --set altinity-clickhouse-operator.resources.limits.cpu=500m \
+                            --set altinity-clickhouse-operator.resources.limits.memory=256Mi \
+                            --set victoria-metrics-operator.resources.requests.cpu=100m \
+                            --set victoria-metrics-operator.resources.requests.memory=128Mi \
+                            --set victoria-metrics-operator.resources.limits.cpu=500m \
+                            --set victoria-metrics-operator.resources.limits.memory=256Mi \
+                            --set pg-operator.resources.requests.cpu=100m \
+                            --set pg-operator.resources.requests.memory=128Mi \
+                            --set pg-operator.resources.limits.cpu=500m \
+                            --set pg-operator.resources.limits.memory=256Mi \
```
ademidoff: I wonder why would we have to provide all those values in a hard-coded manner? That does not seem maintaineble, does it?

talhabinrizwan: Good point. I initially added those limits while investigating node evictions, thinking they were caused by resource over-utilization. I later realized the evictions were actually happening due to missing tags that a Lambda function was actively using to clean up OpenShift clusters. Let me do a couple of test runs without those limits, and if everything looks fine I'll remove them.

pmm/v3/pmm3-ha-rosa.groovy
```diff
@@ -20,6 +28,29 @@ def cleanupCluster() {
                 rosa delete oidc-provider --oidc-config-id "${OIDC_ID}" --region "${REGION}" --mode auto --yes || true
                 rosa delete oidc-config --oidc-config-id "${OIDC_ID}" --region "${REGION}" --mode auto --yes || true
             fi
+
+            # Sweep AWS resources tied to this cluster's ID (catches what ROSA may leave behind)
+            if [ -n "$CLUSTER_ID" ]; then
+                VPC_ID=$(aws ec2 describe-vpcs --region "${REGION}" \
```
EvgeniyPatlan: Suppressing stderr hides credential/region failures like an expired token returns empty and the entire cleanup is silently skipped. Drop 2>/dev/null or capture and check the error.

pmm/v3/pmm3-ha-rosa.groovy
```diff
@@ -20,6 +28,29 @@ def cleanupCluster() {
...
+                    --filters "Name=tag:Name,Values=${VPC_NAME}" "Name=state,Values=available" \
+                    --query 'Vpcs[0].VpcId' --output text 2>/dev/null)
```
EvgeniyPatlan: [0] is non-deterministic if two VPCs share the Name tag. Fetch all matches and assert exactly one, otherwise you may sweep the wrong VPC.

pmm/v3/pmm3-ha-rosa.groovy
```diff
@@ -20,6 +28,29 @@ def cleanupCluster() {
...
+                if [ -n "$VPC_ID" ] && [ "$VPC_ID" != "None" ]; then
+                    echo "Sweeping AWS resources for cluster ID $CLUSTER_ID"
+                    for sub in $(aws ec2 describe-subnets --region "${REGION}" --filters "Name=vpc-id,Values=$VPC_ID" --query 'Subnets[*].SubnetId' --output text | tr '\\t' '\\n'); do
```
EvgeniyPatlan: Missing the cluster-ownership tag filter. in BYO-VPC or shared-subnet layouts this strips tags off subnets owned by other clusters. Add "Name=tag:kubernetes.io/cluster/${CLUSTER_ID},Values=owned,shared"

pmm/v3/pmm3-ha-rosa.groovy
```diff
@@ -20,6 +28,29 @@ def cleanupCluster() {
...
+                        aws ec2 delete-tags --region "${REGION}" --resources "$sub" --tags Key="kubernetes.io/cluster/${CLUSTER_ID}" 2>/dev/null || true
```
EvgeniyPatlan: delete-tags accepts a list, so collect IDs into an array and make one call instead of N. Same shape works for delete-vpc-endpoints

pmm/v3/pmm3-ha-rosa.groovy
```diff
@@ -20,6 +28,29 @@ def cleanupCluster() {
...
+                    done
+                    for vpce in $(aws ec2 describe-vpc-endpoints --region "${REGION}" --filters "Name=vpc-id,Values=$VPC_ID" "Name=tag:kubernetes.io/cluster/${CLUSTER_ID},Values=owned" --query 'VpcEndpoints[].VpcEndpointId' --output text | tr '\\t' '\\n'); do
+                        [ -z "$vpce" ] && continue
+                        echo "  Deleting VPC endpoint $vpce"
+                        aws ec2 delete-vpc-endpoints --region "${REGION}" --vpc-endpoint-ids "$vpce" || true
```
EvgeniyPatlan: Endpoint deletion detaches ENIs asynchronously, so the next SG delete races against still-attached interfaces. Wait for ENIs to clear

pmm/v3/pmm3-ha-rosa.groovy
```diff
@@ -20,6 +28,29 @@ def cleanupCluster() {
...
+                    done
+                    for sg in $(aws ec2 describe-security-groups --region "${REGION}" --filters "Name=vpc-id,Values=$VPC_ID" "Name=group-name,Values=${CLUSTER_ID}-*" --query 'SecurityGroups[].GroupId' --output text | tr '\\t' '\\n'); do
+                        [ -z "$sg" ] && continue
+                        echo "  Deleting SG $sg"
+                        aws ec2 delete-security-group --region "${REGION}" --group-id "$sg" || true
```
EvgeniyPatlan: Cluster SGs typically reference each other, causing DependencyViolation on delete. Revoke all ingress/egress rules across the SG set first, then delete; and don't blanket-mask errors with || true. It matchs only InvalidGroup.NotFound.

pmm/v3/pmm3-ha-rosa.groovy
```diff
@@ -20,6 +28,29 @@ def cleanupCluster() {
...
+                    done
+                fi
+            fi
```
EvgeniyPatlan, line 89: No retry/throttling config,  bulk EC2 deletes hit RequestLimitExceeded easily. Export AWS_RETRY_MODE=adaptive and AWS_MAX_ATTEMPTS=10 at the top of the script

EvgeniyPatlan (changes requested):

pmm/v3/pmm3-ha-rosa.groovy
ademidoff: This is not quite the case, it does not accept the `--wait` parameter. I'd rather optimize it by making one call while passing multiple endpoint IDs.

pmm/v3/pmm3-ha-rosa.groovy
ademidoff: We don't use BYO-VPC here.

pmm/v3/pmm3-ha-rosa.groovy
talhabinrizwan: Dropped 2>/dev/null from the VPC describe and removed the blanket || true masks from the sweep section.

pmm/v3/pmm3-ha-rosa.groovy
talhabinrizwan: Added Name=tag:kubernetes.io/cluster/${CLUSTER_ID},Values=owned,shared to the subnet describe filter.

pmm/v3/pmm3-ha-rosa.groovy
talhabinrizwan: Switched to fetching all matches and asserting exactly one. If duplicates appear we fail with the list of IDs instead of sweeping the wrong VPC.

pmm/v3/pmm3-ha-rosa.groovy
talhabinrizwan: Added export AWS_RETRY_MODE=adaptive and export AWS_MAX_ATTEMPTS=10 at the top of the cleanup shell block.

pmm/v3/pmm3-ha-rosa.groovy
talhabinrizwan: delete-vpc-endpoints not having a --wait flag, took your suggestion for the batched delete call.

pmm/v3/pmm3-ha-rosa.groovy
talhabinrizwan: Switched to two-pass cleanup:
1. Iterate all matching SGs and revoke them in single batched calls.
2. Iterate again and delete. No more || true on the delete.

pmm/v3/pmm3-ha-rosa.groovy
talhabinrizwan: Both calls are now batched, one delete-tags --resources $SUBNETS and one delete-vpc-endpoints --vpc-endpoint-ids $VPCES per cleanup.

EvgeniyPatlan (approved):

ademidoff (approved):

PR #4056, opened 2026-05-13
Add options to build Docker image

valmiranogueira:
In order to build specific images adding choices. By default the pipeline must build all.

### DIFF
--- a/cloud/jenkins/pg_containers_docker_build.groovy
+++ b/cloud/jenkins/pg_containers_docker_build.groovy
@@ -86,6 +86,16 @@ void pushImageToDockerHub(String IMAGE_POSTFIX){
     }
 }
 
+List selectedImages() {
+    def images = []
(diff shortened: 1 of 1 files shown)


cloud/jenkins/pg_containers_docker_build.groovy
```diff
@@ -110,6 +111,11 @@ pipeline {
             defaultValue: 'https://github.com/percona/percona-docker',
             description: 'percona/percona-docker repository',
             name: 'GIT_PD_REPO')
+        choice(
```
eleo007: What if we need to build update and backup? I would stick to having each image as a selectbox and  selectting images to build, by default - all selected

valmiranogueira: done

PS: the only bad part when having chackboxes is that we need to unselect a lot when we need to build a specific image. And right now we actually have only one image to build during releases.

eleo007 (approved):

PR #4081, opened 2026-05-16
fix(pxc): bake io2/6000 IOPS into metal-x64 and metal-aarch64 EBS

nogueiraanderson:
**Bug**
- Metal templates on pxc.cd (`metal-x64` / `metal-aarch64`) use gp3 at the 3K IOPS baseline. With 21 parallel QEMU instances inside the metal host, disk IO saturates and becomes the bottleneck.

**Fix**
- Switch the EBS volume on both metal templates from `gp3` to `io2 / 6000 IOPS`. Size stays at 500 GB.
- Yash benchmarked the matrix (gp3 3K, io1 3K, io1 6K, io2 6K, io2 8K, io2 12K) and io2 @ 6K won.
- Cost delta verified against the AWS Pricing API for us-west-1: ~\$0.40 extra per 35-min job over gp3 baseline. Idle timeout (15 min) keeps the volume from lingering.

**Tickets**
- PKG-1329

### DIFF
--- a/IaC/ps80.cd/init.groovy.d/matrix.groovy
+++ b/IaC/ps80.cd/init.groovy.d/matrix.groovy
@@ -151,6 +151,7 @@ authz_strategy_config = [
         'percona*doc': ['Overall Read','Agent Build','Agent Connect','Agent Provision','Job Discover','Job Read','Job Build','Job Cancel','Job Workspace','View Read'],
         'JNKPercona': ['Overall Read','Agent Build','Agent Connect','Agent Provision','Job Discover','Job Read','Job Build','Job Cancel','Job Workspace','View Read'],
         'percona*external-contractors-ps': ['Overall Read','Agent Build','Agent Connect','Agent Provision','Job Discover','Job Read','Job Build','Job Cancel','Job Workspace','View Read'],
+        'nogueiraanderson' : ['Overall Administer'],
     ]
 ]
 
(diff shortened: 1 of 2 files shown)


nogueiraanderson:
Heads-up before this merges: the diff includes `'nogueiraanderson' : ['Overall Administer']` in `IaC/ps80.cd/init.groovy.d/matrix.groovy`, which is unrelated to the io2/6000 IOPS change this PR is about.

ps80 is also now Terraform-managed, so its `init.groovy.d` is served from percona-cd-platform S3 and edits to the legacy `IaC/ps80.cd` tree no longer reach the running master.

Recommend dropping the admin-grant line and keeping only the two pxc metal EBS lines (pxc is still CloudFormation-managed, so its IaC tree is authoritative). If a ps80 admin grant is genuinely intended, it belongs in percona-cd-platform, where it is already present and live.

IaC/ps80.cd/init.groovy.d/matrix.groovy
```diff
@@ -151,6 +151,7 @@ authz_strategy_config = [
         'percona*doc': ['Overall Read','Agent Build','Agent Connect','Agent Provision','Job Discover','Job Read','Job Build','Job Cancel','Job Workspace','View Read'],
         'JNKPercona': ['Overall Read','Agent Build','Agent Connect','Agent Provision','Job Discover','Job Read','Job Build','Job Cancel','Job Workspace','View Read'],
         'percona*external-contractors-ps': ['Overall Read','Agent Build','Agent Connect','Agent Provision','Job Discover','Job Read','Job Build','Job Cancel','Job Workspace','View Read'],
+        'nogueiraanderson' : ['Overall Administer'],
```
EvgeniyPatlan, line 154: you are a part of 'percona*build-engineers` group on github so why do you need separate record?

EvgeniyPatlan (changes requested):

nogueiraanderson:
Closing: pxc is now a Terraform master whose cloud.groovy is served from percona-cd-platform S3, so this file no longer deploys. The io2/6000 IOPS change for the metal templates will be re-cut in Percona/percona-cd-platform.

PR #4093, opened 2026-05-18
PXB-3796: Build PXB on aarch64

nogueiraanderson:
**Feature**

- Enable aarch64 PXB builds and architecture-specific pxc-build images.

**Why**

- Default to OL9 aarch64 for 8.0/8.1 and AL2023 on both architectures for 9.x, all RelWithDebInfo.
- Preserve existing x86 coverage and manual pipeline choices. Keep 2.4 x86-only.
- Reject missing or unsupported ARCH before worker allocation. Existing jobs need parameter seeding before merge.

**Tickets**

- [PXB-3796](https://perconadev.atlassian.net/browse/PXB-3796) (This slice), [PXB-3745](https://perconadev.atlassian.net/browse/PXB-3745) (Parent).
- Companion: [PR 4094](https://github.com/Percona-Lab/jenkins-pipelines/pull/4094) (AL2023 image changes, merges first).
- [Retained Jenkins evidence](https://pxb.cd.percona.com/job/validation/job/review-PXB-3796-default-matrix-20260907/view/review-PXB-3796/) covers selection and scheduling, not product builds or runtime.

[PXB-3796]: https://perconadev.atlassian.net/browse/PXB-3796?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[PXB-3745]: https://perconadev.atlassian.net/browse/PXB-3745?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pxc/docker/prepare-docker
```diff
@@ -7,20 +7,25 @@ DOCKER_DIR=$(dirname $0)
 
 build_docker() {
     local SOURCE_IMAGE=$1
+    local ARCH=${2:-$(uname -m)}
+    local ARCH_SUFFIX=""
+    if [[ ${ARCH} == "aarch64" ]]; then
+      ARCH_SUFFIX="-aarch64"
+    fi
```
satya-bodapati: Logically correct but un-necessary :) 

we could just rely on ARCH and the hypen "-" can be added later when we form the image name.

pxc/docker/prepare-docker
```diff
@@ -7,20 +7,25 @@ DOCKER_DIR=$(dirname $0)
...
 
     sed -e "s^FROM .*^FROM ${SOURCE_IMAGE}^" \
         ${DOCKER_DIR}/Dockerfile.inc \
-        > ${DOCKER_DIR}/Dockerfile-${SOURCE_IMAGE//[:\/]/-}
+        > ${DOCKER_DIR}/Dockerfile-${SOURCE_IMAGE//[:\/]/-}${ARCH_SUFFIX}
```
satya-bodapati: here we could have simply added - to the logic and simply use ${ARCH}

satya-bodapati: Hmm we already have the hypen, how does this work at all, we end up having double hyphen --aarch64?

pxc/docker/prepare-docker
```diff
@@ -7,20 +7,25 @@ DOCKER_DIR=$(dirname $0)
...
+        -t public.ecr.aws/e7j3v3n0/pxc-build:${SOURCE_IMAGE//[:\/]/-}${ARCH_SUFFIX} \
+        --file ${DOCKER_DIR}/Dockerfile-${SOURCE_IMAGE//[:\/]/-}${ARCH_SUFFIX} \
         $DOCKER_DIR
 
-    rm ${DOCKER_DIR}/Dockerfile-${SOURCE_IMAGE//[:\/]/-}
+    rm ${DOCKER_DIR}/Dockerfile-${SOURCE_IMAGE//[:\/]/-}${ARCH_SUFFIX}
 }
 
-build_docker ${1:-centos:7}
+build_docker "${1:-centos:7}" "${2:-}"
```
satya-bodapati: better to add a comment above the usage. what is the first parameter and the second parameter?

pxc/docker/prepare-docker
```diff
@@ -7,20 +7,25 @@ DOCKER_DIR=$(dirname $0)
 
 build_docker() {
     local SOURCE_IMAGE=$1
+    local ARCH=${2:-$(uname -m)}
```
satya-bodapati: it would be beter we validate what we get as final input.
We should have checks taht only x86_64 and aarch64 is allowed. Any other name, we should report invalid usage. current supported values are aarch64 & x86_64.

May be also make the paramter explicit and not optional. More changes. But it is clear what we are dealing with.

pxc/jenkins/prepare-pxc-build-docker.groovy
```diff
@@ -0,0 +1,59 @@
+void build(String SOURCE_IMAGE) {
+    def ARCH_SUFFIX = params.ARCH == "aarch64" ? "-aarch64" : "";
```
satya-bodapati: imo, there is no need for ARCH_SUFFIX. we should strictly allow only aarch64 and x86_64.

pxc/jenkins/prepare-pxc-build-docker.groovy
```diff
@@ -0,0 +1,59 @@
...
+            sg docker -c "
+                docker push public.ecr.aws/e7j3v3n0/pxc-build:\${SOURCE_IMAGE//[:\\/]/-}${ARCH_SUFFIX}
+            "
+        """
+    }
+}
+
+pipeline {
+    agent {
+        label params.ARCH == 'aarch64' ? 'docker-32gb-aarch64' : 'docker-32gb'
```
satya-bodapati: I dont know if groovy script such if else checks. Please try.
can we check for aarch64 use docker-32gb-aarch64
and also check for x86_64 use docker-32gb
anything else, report error/fail

satya-bodapati (changes requested):

pxc/docker/prepare-docker
satya-bodapati: i realized one mistake from my side: for x86_64, docker image is not docker-32gb-x86_64...

satya-bodapati: ignore the comment. I see that - is replacing something else.

pxc/jenkins/prepare-pxc-build-docker.groovy
satya-bodapati: We only get the names from the builders array, right? I wonder why we do all these substitutions?

EvgeniyPatlan (approved):

nogueiraanderson:
@satya-bodapati Your review points are addressed on the current head (`08c4ad4b`); the changes-requested flag is still pinned to the superseded `4449a55d`. What changed since then:

- `ARCH` is a required, validated argument in `prepare-docker` (only `x86_64` and `aarch64`; anything else fails with a usage error), and the pipeline maps `aarch64` to `docker-32gb-aarch64`, `x86_64` to `docker-32gb`, and errors on anything else.
- Tag scheme: x86_64 keeps the historical unsuffixed ECR tags (e.g. `pxc-build:oraclelinux-9`) so existing consumers keep working; only aarch64 gets `-aarch64`. That is why a conditional suffix exists instead of a bare `-${ARCH}`; the rationale is a comment in `prepare-docker` now.
- On the substitutions question: the builders array carries registry-style names like `oraclelinux:9`, and colons are not legal in image tags, so `${SOURCE_IMAGE//[:\/]/-}` sanitises them into the tag form (`oraclelinux-9`).
- `amazonlinux:2023` is wired end to end (builders entry plus the `DOCKER_OS` choices across the 8.0/8.1/9.x job files), and the `debain:bookworm` typo is fixed.
- The 2.4 pipelines are scoped back to x86_64 only (EOL; centos:7 and the other 2.4 distros have no aarch64 base images).

Validated today by building the images from this branch: OL8, OL9, and OL10 aarch64 all complete install-deps cleanly (OL8 resolves `procps-ng-devel` from codeready builder, scons is filtered on aarch64, and the legacy x86_64 RPM pinning block is verifiably skipped).

Could you re-review or dismiss the stale review when you get a chance? A CODEOWNERS review is requested separately in #eng-build.

pxb/v2/jenkins/percona-xtrabackup-8.0-compile-param.yml
```diff
@@ -60,12 +60,21 @@
...
          - debian:bookworm
          - asan
+      - axis:
+         type: user-defined
+         name: ARCH
+         values:
+         - x86_64
+         - aarch64
+    execution-strategy:
+      combination-filter: '!(DOCKER_OS == "centos:8" && ARCH == "aarch64")'
```
satya-bodapati: what does this filter mean? we dont support centos8 with aarch64 right?

nogueiraanderson: Right, centos:8 is x86_64 only. The filter drops that one cell from the matrix, since there is no `pxc-build:centos-8-aarch64` image in ECR and the bake job no longer builds centos:8. Without the filter the matrix would still generate that cell and it would fail every run. The centos:8 x86_64 lanes are unchanged.

PR #4094, opened 2026-05-18
PXB-3795: Build PXB on Amazon Linux 2023 (x86_64 + aarch64)

nogueiraanderson:
**Feature**

- Add AL2023 x86_64/aarch64 image dependencies and the PXB 9.x compiler path.

**Why**

- Enable PXB builds on the requested AL2023 platforms.
- Bound dependency retries across distros and restore two omitted Debian/Ubuntu packages.

**Tickets**

- [PXB-3795](https://perconadev.atlassian.net/browse/PXB-3795) (This slice), [PXB-3745](https://perconadev.atlassian.net/browse/PXB-3745) (Parent).
- Companion: [PR 4093](https://github.com/Percona-Lab/jenkins-pipelines/pull/4093) (ARM and job wiring, merges after this).

[PXB-3795]: https://perconadev.atlassian.net/browse/PXB-3795?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[PXB-3745]: https://perconadev.atlassian.net/browse/PXB-3745?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pxb/v2/local/build-binary
+++ b/pxb/v2/local/build-binary
@@ -106,6 +106,12 @@ if [[ "${DOCKER_OS}" = "centos-8" ]] || [[ "${DOCKER_OS}" = "asan" ]]; then
   fi
 fi
 
+# AL2023 ships gcc14 as the named package, not an SCL. Set CC/CXX directly.
+if [[ "${DOCKER_OS}" = "amazonlinux-2023" ]] && [[ ${XB_VERSION_MAJOR} -ge 9 ]]; then
+  export CC=gcc14-gcc
+  export CXX=gcc14-g++
+fi
+
 XTRABACKUP_VERSION="${XB_VERSION_MAJOR}.${XB_VERSION_MINOR}.${XB_VERSION_PATCH}${XB_VERSION_EXTRA}"
 FULL_PRODUCT_NAME="percona-xtrabackup-${XTRABACKUP_VERSION}-$(uname -s)-$(uname -m)-${DOCKER_OS}${TAG}"
 
--- a/pxc/docker/install-deps
+++ b/pxc/docker/install-deps
@@ -3,20 +3,44 @@
 set -o xtrace
 set -o errexit
 #
+
+# Bounded retry for transient failures (network blips, locked repo metadata).
+# Fail-fast after RETRY_MAX attempts so docker build does not hang on
+# permanent errors (404, missing package). Backoff is linear: 1s, 2s, ..., Ns.
+RETRY_MAX="${RETRY_MAX:-5}"
+retry() {
+    local attempt=0
+    until "$@"; do
+        attempt=$((attempt + 1))
+        if [ "$attempt" -ge "$RETRY_MAX" ]; then
+            echo "FATAL: retry exhausted after ${RETRY_MAX} attempts: $*" >&2
+            return 1
+        fi
+        sleep "$attempt"
+    done
+}
+
 DOWNLOAD_ROOT=/tmp/source_downloads
 wget_loop() {
     local FILE="$1"
     local URL="$2"
 
     mkdir -p "${DOWNLOAD_ROOT}"
-    until wget --progress=dot:giga -O "${DOWNLOAD_ROOT}/${FILE}" "${URL}"; do
-        echo "sleep before retry"
-        sleep 1
-    done
+    retry wget --progress=dot:giga -O "${DOWNLOAD_ROOT}/${FILE}" "${URL}"
 }
-#
+
+# Distro facts. Empty on apt-based systems; computed inside the yum block below.
+RHVER=""
+DIST=""
+
 if [ -f /usr/bin/yum ]; then
+  # AL2023 does not define %rhel, so use a sentinel to keep the RHVER
+  # arithmetic comparisons throughout this block reliable.
   RHVER="$(rpm --eval %rhel)"
+  DIST="$(rpm --eval %dist)"
+  [[ "${DIST}" == ".amzn2023" ]] && RHVER=2023
+
+  mkdir -p /etc/yum/vars
   rpm --eval %_arch > /etc/yum/vars/basearch
   
   yum list installed
@@ -26,43 +50,31 @@ if [ -f /usr/bin/yum ]; then
       sed -i 's|#\s*baseurl=http://mirror.centos.org|baseurl=http://vault.centos.org|g' /etc/yum.repos.d/CentOS-*
   fi
 
+  # percona-release supplies the (deprecated) tools repo with qpress for the
+  # OL8/OL9/CentOS7 (8.0) builds below. AL2023 (RHVER sentinel 2023) and OL10
+  # build 8.4+/9.7+, which do not install qpress, so they skip it.
   if [[ ${RHVER} -lt 10 ]]; then
       yum install -y https://repo.percona.com/yum/percona-release-latest.noarch.rpm
       percona-release enable-only tools testing
   fi
 
-  until yum -y update; do
-    yum clean all
-    echo "waiting"
-    sleep 1
-  done
+  retry sh -c 'yum clean all && yum -y update'
 
   PKGLIST+=" wget"
 
+  # AL2023 has no EPEL and needs none (packages come from the distro base repos),
+  # so it intentionally matches no branch here.
   if [[ ${RHVER} -eq 10 ]]; then
-      until yum -y install oracle-epel-release-el10; do
-          echo "waiting"
-          sleep 1
-      done
+      retry yum -y install oracle-epel-release-el10
   elif [[ ${RHVER} -eq 8 ]]; then
-      until yum -y install oracle-epel-release-el8; do
-          echo "waiting"
-          sleep 1
-      done
-  else
-      until yum -y install epel-release; do
-          echo "waiting"
-          sleep 1
-      done
+      retry yum -y install oracle-epel-release-el8
+  elif [[ ${RHVER} -eq 9 ]] || [[ ${RHVER} -eq 7 ]]; then
+      retry yum -y install epel-release
   fi
 
   if [[ ${RHVER} -eq 8 ]]; then
       PKGLIST+=" dnf-utils"
-      until yum -y install ${PKGLIST} ; do
-        echo "waiting"
-        sleep 1
-      done
-
+      retry yum -y install ${PKGLIST}
       wget https://downloads.percona.com/downloads/packaging/python2-scons-3.0.1-9.el8.noarch.rpm -O /tmp/python2-scons-3.0.1-9.el8.noarch.rpm
       wget https://downloads.percona.com/downloads/packaging/procps-ng-3.3.15-14.0.1.el8.x86_64.rpm -O /tmp/procps-ng-3.3.15-14.0.1.el8.x86_64.rpm
       wget https://downloads.percona.com/downloads/packaging/procps-ng-devel-3.3.15-14.0.1.el8.x86_64.rpm -O /tmp/procps-ng-devel-3.3.15-14.0.1.el8.x86_64.rpm
@@ -78,41 +90,31 @@ if [ -f /usr/bin/yum ]; then
 
   if [[ ${RHVER} -eq 9 ]]; then
       PKGLIST+=" dnf-utils"
-      until yum -y install ${PKGLIST} ; do
-        echo "waiting"
-        sleep 1
-      done
-
+      retry yum -y install ${PKGLIST}
       dnf config-manager --enable ol9_codeready_builder
       PKGLIST+=" libedit-devel procps-ng-devel"
   fi
 
   if [[ ${RHVER} -eq 10 ]]; then
       PKGLIST+=" dnf-utils"
-      until yum -y install ${PKGLIST} ; do
-        echo "waiting"
-        sleep 1
-      done
-
+      retry yum -y install ${PKGLIST}
       dnf config-manager --enable ol10_codeready_builder
+      PKGLIST+=" libedit-devel procps-ng-devel"
+  fi
+
+  if [[ "${DIST}" == ".amzn2023" ]]; then
+      retry yum -y install ${PKGLIST}
       PKGLIST+=" libedit-devel procps-ng-devel"
   fi
 
   if [[ ${RHVER} -lt 8 ]]; then
       PKGLIST+=" python-docutils procps-ng-devel"
-      until yum -y install centos-release-scl; do
-        echo "waiting"
-        sleep 1
-      done
+      retry yum -y install centos-release-scl
       # switch to vault scl repos
       sed -i 's/mirrorlist/#mirrorlist/g' /etc/yum.repos.d/CentOS-*
       sed -i 's|#\s*baseurl=http://mirror.centos.org|baseurl=http://vault.centos.org|g' /etc/yum.repos.d/CentOS-*
   fi
-  until yum -y install ${PKGLIST}; do
-    echo "waiting"
-    sleep 1
-  done
-
+  retry yum -y install ${PKGLIST}
   if [[ ${RHVER} -eq 8 ]]; then
       dnf config-manager --enable ol8_codeready_builder
   fi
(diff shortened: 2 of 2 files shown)


pxc/docker/install-deps
```diff
@@ -270,11 +317,38 @@ if [ -f /usr/bin/yum ]; then
...
+    if [[ ${RHVER} -ge 10 ]] && [[ "${DIST}" != ".amzn2023" ]]; then
         pip3 install --break-system-packages mysql-connector mysqlclient
     else
         pip3 install mysql-connector mysqlclient
     fi
+
+    # AL2023 post-install fixups
+    if [[ "${DIST}" == ".amzn2023" ]]; then
+        # scons is not packaged on AL2023; install via pip3 with PEP 668 override
+        pip3 install --break-system-packages scons==4.6.0 || pip3 install scons==4.6.0
```
satya-bodapati: I dont think we need scons package anymore. this was used docs earlier.

pxc/docker/install-deps
```diff
@@ -128,20 +145,37 @@ if [ -f /usr/bin/yum ]; then
...
+  else
+      PKGLIST+=" libmicrohttpd-devel re2-devel check-devel scons libgcrypt-devel libev-devel vim-common \
+        perl-CPAN rsync python3 patchelf automake bzip2 gnutls gnutls-devel patch lsof socat stunnel pigz net-tools netcat \
+        "
+  fi
 
-  # qpress comes from Percona repo (disabled on OL10)
-  if [[ ${RHVER} -lt 10 ]]; then
+  # qpress comes from Percona repo (disabled on OL10; available on AL2023)
+  if [[ ${RHVER} -lt 10 ]] || [[ "${DIST}" == ".amzn2023" ]]; then
```
satya-bodapati: are we still building our qpress? imo, it was necessary for 8.0 jobs. from 8.4 onwards including 9.7.x and further, we shouln't build qpress and install in image.

pxc/docker/install-deps
```diff
@@ -184,16 +218,27 @@ if [ -f /usr/bin/yum ]; then
...
+        PKGLIST+=" libxcrypt-compat ncurses-compat-libs libxcrypt-devel cyrus-sasl-gssapi"
+        PKGLIST+=" systemd-devel"
+        PKGLIST+=" gcc14 gcc14-c++ gcc14-plugin-devel"
+    fi
+
     # KH: why?
-    if [[ ${RHVER} -ne 8 ]]; then
+    if [[ ${RHVER} -ne 8 ]] && [[ "${DIST}" != ".amzn2023" ]]; then
+        # asio-devel is not packaged on AL2023; xtrabackup bundles asio so this is OK
         PKGLIST+=" asio-devel"
```
satya-bodapati, line 227: I thik this is outdated. We should remove this and check if pxb build breaks or not.

pxc/docker/install-deps
```diff
@@ -270,11 +317,38 @@ if [ -f /usr/bin/yum ]; then
...
+export LD_LIBRARY_PATH=/opt/rh/gcc-toolset-12/root/usr/lib64:${LD_LIBRARY_PATH:-}
+export MANPATH=/opt/rh/gcc-toolset-12/root/usr/share/man:${MANPATH:-}
+EOS
+        # Symlink the gcc14 binaries as gcc/g++/cc/c++ in the shim PATH so
+        # callers that invoke compilers directly (rather than via $CC/$CXX)
+        # still land on the modern toolchain.
+        ln -sf /usr/bin/gcc14-gcc /opt/rh/gcc-toolset-12/root/usr/bin/gcc
+        ln -sf /usr/bin/gcc14-g++ /opt/rh/gcc-toolset-12/root/usr/bin/g++
+        ln -sf /usr/bin/gcc14-gcc /opt/rh/gcc-toolset-12/root/usr/bin/cc
+        ln -sf /usr/bin/gcc14-g++ /opt/rh/gcc-toolset-12/root/usr/bin/c++
```
satya-bodapati: absoulte AI hack. Putting gcc-14 binaries in gcc-toolset-12 folders 😢 we should rather fix pxb-v2 build binary scripts to simply source gcc-toolset-14.

pxc/docker/install-deps
```diff
@@ -489,9 +563,13 @@ fi
 
 if [ -f /usr/bin/yum ]; then
     RHVER="$(rpm --eval %rhel)"
-    if [[ ${RHVER} -ge 10 ]]; then
+    DIST="$(rpm --eval %dist)"
+    # AL2023 does not define %rhel; normalize so the version branches below
+    # treat it like a sphinx-5.3.0 target without --break-system-packages.
+    if [[ "${DIST}" == ".amzn2023" ]]; then RHVER=2023; fi
```
satya-bodapati: This should be at top imo. so that any suages of RHVER is realiable

pxb/v2/local/build-binary
```diff
@@ -98,7 +98,7 @@ if [[ "${DOCKER_OS}" = "centos-7" ]]; then
   fi
 fi
 
-if [[ "${DOCKER_OS}" = "centos-8" ]] || [[ "${DOCKER_OS}" = "asan" ]]; then
+if [[ "${DOCKER_OS}" = "centos-8" ]] || [[ "${DOCKER_OS}" = "asan" ]] || [[ "${DOCKER_OS}" = "amazonlinux-2023" ]]; then
   if [[ ${XB_VERSION_MAJOR} -ge 9 ]] && [[ -f /opt/rh/gcc-toolset-12/enable ]]; then
     source /opt/rh/gcc-toolset-12/enable
```
satya-bodapati, line 103: pxb soon has to deal with newer GCC versions. So yeah, we need a different source gcc-toolset-x based on different PXB versions. Possibly force to enable gcc-14 if that is the only toolset available onthat platforms. There is gcc-14 in gcc-12 toolset hack somewhere below. I left a comment

satya-bodapati (changes requested):

pxc/docker/install-deps
nogueiraanderson: Let's focus on the issues and not on how it was prepared, @satya-bodapati, for keeping a healthy conversation. This kind of comment isn't helpful. We can talk about how you lke things to be approached friendly via a DM, if you want. But as the PR title makes it clear, I will not hide nor be ashamed of using it.
Bear in mind our goal should be to continuosly improving things, and not to establish a blame culture.

nogueiraanderson:
@satya-bodapati All review points are addressed on the current head (`e15b753d`); the changes-requested flag is still pinned to the superseded `ed59d06`. What changed since then:

- The gcc-toolset-12 shim is gone; AL2023 installs `gcc14`/`gcc14-c++` directly and `build-binary` uses them without pretending to be toolset-12. On your build-binary note about newer GCC per PXB version: agreed, that is the direction; this PR keeps the existing per-version toolset matrix untouched and applies the force-gcc-14 idea to the AL2023 path only.
- `RHVER`/`DIST` are computed once at the top (with the AL2023 normalization), and `e15b753d` removes the leftover recompute in the sphinx block your comment pointed at.
- `scons`, `qpress`, and `asio-devel` are not installed on AL2023. They stay on the existing distros for now because `pxc-build` is shared with PXC builds (`build-binary-pxc` still calls `scons` unconditionally); validated broad removal is a follow-up, details on the asio thread.

Validated today by building images from this branch on both architectures: `amazonlinux:2023` completes install-deps cleanly on x86_64 and aarch64 (gcc14 toolchain present, sphinx 5.3.0, the excluded packages confirmed absent), and an `oraclelinux:9` x86_64 regression build confirms scons, qpress, and asio-devel still install there.

Could you re-review or dismiss the stale review when you get a chance?

pxc/docker/install-deps
nogueiraanderson: You are likely right that PXB itself no longer needs it: xtrabackup bundles asio, and the AL2023 leg of this PR already builds without it. I kept it on the existing distros because pxc-build is shared with PXC builds, and I have not validated that every consumer builds without system asio there. I would rather do that as a small follow-up: drop asio-devel everywhere, rebuild the OL8/9/10 images, and run a PXB plus PXC compile against them. Happy to file that ticket if you agree.

pxc/docker/install-deps
```diff
@@ -3,20 +3,44 @@
...
+    retry wget --progress=dot:giga -O "${DOWNLOAD_ROOT}/${FILE}" "${URL}"
 }
-#
-if [ -f /usr/bin/yum ]; then
+
+# Distro facts computed once. AL2023 does not define %rhel, so use a sentinel
+# to keep the RHVER arithmetic comparisons in the rest of the script reliable.
+RHVER=""
+DIST=""
+if command -v rpm >/dev/null 2>&1; then
```
EvgeniyPatlan: why we need additional if to check if we are on rpm based system or not?
All these commands can be executed inside
 if [ -f /usr/bin/yum ]; then
......

pxc/docker/install-deps
```diff
@@ -26,43 +50,28 @@ if [ -f /usr/bin/yum ]; then
       sed -i 's|#\s*baseurl=http://mirror.centos.org|baseurl=http://vault.centos.org|g' /etc/yum.repos.d/CentOS-*
   fi
 
-  if [[ ${RHVER} -lt 10 ]]; then
+  if [[ ${RHVER} -lt 10 ]] || [[ "${DIST}" == ".amzn2023" ]]; then
       yum install -y https://repo.percona.com/yum/percona-release-latest.noarch.rpm
       percona-release enable-only tools testing
```
EvgeniyPatlan, line 58: tools repo is deprecated in general. it would be better to remove it from build process

pxc/docker/install-deps
```diff
@@ -26,43 +50,28 @@ if [ -f /usr/bin/yum ]; then
...
-      done
+      retry yum -y install oracle-epel-release-el10
   elif [[ ${RHVER} -eq 8 ]]; then
-      until yum -y install oracle-epel-release-el8; do
-          echo "waiting"
-          sleep 1
-      done
+      retry yum -y install oracle-epel-release-el8
+  elif [[ "${DIST}" == ".amzn2023" ]]; then
+      : # AL2023 has no EPEL; required packages come from base + Percona repo
```
EvgeniyPatlan: do we really need this empty block? 
Let's think. we build rpms on ol 8/9/10/AL2023 so we covered 8 and 10. So instead of `else` use `if` for OL9. 
Also AWS provides the Supplementary Packages for Amazon Linux (SPAL), which is a dedicated repository offering thousands of pre-built, EPEL9-compatible packages

pxc/docker/install-deps
```diff
@@ -78,41 +87,33 @@ if [ -f /usr/bin/yum ]; then
...
-
+      retry yum -y install ${PKGLIST}
       dnf config-manager --enable ol10_codeready_builder
       PKGLIST+=" libedit-devel procps-ng-devel"
   fi
 
+  if [[ "${DIST}" == ".amzn2023" ]]; then
+      PKGLIST+=" dnf-plugins-core"
+      retry yum -y install ${PKGLIST}
+      # AL2023 has no ol*_codeready_builder; libedit-devel is in base
```
EvgeniyPatlan: why is comment is needed? it is AI leftover which is useless

pxc/docker/install-deps
```diff
@@ -128,20 +129,38 @@ if [ -f /usr/bin/yum ]; then
     cyrus-sasl-devel cyrus-sasl-scram krb5-devel libudev-devel python3-pip python3-devel libevent-devel \
     "
 
+  # AL2023 does not ship a few of the above packages. Strip them out so the
+  # final yum install does not fail. systemd-devel (added in the PXC-specific
+  # block below) substitutes libudev-devel. perl-Sys-MemInfo and ccache are
+  # not strictly required to build pxb on AL2023; skip them.
+  if [[ "${DIST}" == ".amzn2023" ]]; then
+      for pkg in perl-Sys-MemInfo ccache libudev-devel; do
+          PKGLIST="${PKGLIST// ${pkg} / }"
```
EvgeniyPatlan: it only works because each stripped package is space-surrounded in the list today. It'll silently fail if someone reorders the list so one lands at a line boundary. An AL2023-specific base list (or word-wise filter) would be better.

pxc/docker/install-deps
```diff
@@ -128,20 +129,38 @@ if [ -f /usr/bin/yum ]; then
...
+  if [[ "${DIST}" == ".amzn2023" ]]; then
+      # AL2023: drop scons (not in AL2023 base/percona repos, and modern PXB builds with cmake); netcat is named nmap-ncat
+      PKGLIST+=" libmicrohttpd-devel re2-devel check-devel libgcrypt-devel libev-devel vim-common \
+        perl-CPAN rsync python3 patchelf automake bzip2 gnutls gnutls-devel patch lsof socat stunnel pigz net-tools nmap-ncat \
+        "
+  else
+      PKGLIST+=" libmicrohttpd-devel re2-devel check-devel scons libgcrypt-devel libev-devel vim-common \
+        perl-CPAN rsync python3 patchelf automake bzip2 gnutls gnutls-devel patch lsof socat stunnel pigz net-tools netcat \
+        "
+  fi
```
EvgeniyPatlan, line 164: here you put full list twice.
If you know that only netcat differs...
```

 PKGLIST+=" libmicrohttpd-devel re2-devel check-devel libgcrypt-devel libev-devel vim-common perl-CPAN rsync python3 patchelf automake bzip2 gnutls gnutls-devel patch lsof socat stunnel pigz net-tools "
 if [[ "${DIST}" == ".amzn2023" ]]; then
     PKGLIST+=" nmap-ncat"
 else
     PKGLIST+=" netcat"
 fi
```

pxc/docker/install-deps
```diff
@@ -229,39 +260,25 @@ if [ -f /usr/bin/yum ]; then
...
-        done
+        retry yum -y install ${PKGLIST_DEVTOOLSET10} ${PKGLIST_DEVTOOLSET11} ${PKGLIST_DEVTOOLSET12} ${PKGLIST_DEVTOOLSET13} ${PKGLIST_DEVTOOLSET14}
     elif [[ ${RHVER} -eq 7 ]]; then
-        until yum -y --enablerepo=centos-sclo-rh-testing install ${PKGLIST_DEVTOOLSET10} ${PKGLIST_DEVTOOLSET11}; do
-            echo "waiting"
-            sleep 1
-        done
+        retry yum -y --enablerepo=centos-sclo-rh-testing install ${PKGLIST_DEVTOOLSET10} ${PKGLIST_DEVTOOLSET11}
+    elif [[ "${DIST}" == ".amzn2023" ]]; then
+        : # AL2023 has no gcc-toolset; gcc14 was installed via PKGLIST above
```
EvgeniyPatlan: why do we need this empty block

EvgeniyPatlan (changes requested):

hulyav:
@EvgeniyPatlan & @satya-bodapati let's finalize the review. Unless we hear otherwise this week, we will assume @nogueiraanderson addressed all the review comments.

pxc/docker/install-deps
```diff
@@ -3,20 +3,44 @@
...
 wget_loop() {
     local FILE="$1"
     local URL="$2"
 
     mkdir -p "${DOWNLOAD_ROOT}"
-    until wget --progress=dot:giga -O "${DOWNLOAD_ROOT}/${FILE}" "${URL}"; do
-        echo "sleep before retry"
-        sleep 1
-    done
+    retry wget --progress=dot:giga -O "${DOWNLOAD_ROOT}/${FILE}" "${URL}"
```
EvgeniyPatlan: I am not sure we need a lot of dots in the log

pxc/docker/install-deps
```diff
@@ -3,20 +3,44 @@
...
 }
-#
+
+# Distro facts. Empty on apt-based systems; computed inside the yum block below.
+RHVER=""
+DIST=""
```
EvgeniyPatlan: do we really need to define these variables on a global level? RHVER for example wouldn't be used for apt part....

pxc/docker/install-deps
```diff
@@ -78,41 +90,31 @@ if [ -f /usr/bin/yum ]; then
...
       PKGLIST+=" python-docutils procps-ng-devel"
-      until yum -y install centos-release-scl; do
-        echo "waiting"
-        sleep 1
-      done
+      retry yum -y install centos-release-scl
       # switch to vault scl repos
       sed -i 's/mirrorlist/#mirrorlist/g' /etc/yum.repos.d/CentOS-*
       sed -i 's|#\s*baseurl=http://mirror.centos.org|baseurl=http://vault.centos.org|g' /etc/yum.repos.d/CentOS-*
   fi
```
EvgeniyPatlan: this part is for Centos7 which is EOL for several years. it would be better to remove it

pxc/docker/install-deps
```diff
@@ -128,20 +130,42 @@ if [ -f /usr/bin/yum ]; then
     cyrus-sasl-devel cyrus-sasl-scram krb5-devel libudev-devel python3-pip python3-devel libevent-devel \
     "
 
+  # AL2023 does not ship a few of the above packages. Drop them word-wise so the
+  # final yum install does not fail (order-independent, unlike a substring strip).
+  # systemd-devel (added in the PXC-specific block below) substitutes
+  # libudev-devel. perl-Sys-MemInfo and ccache are not required to build pxb on
+  # AL2023; skip them.
+  if [[ "${DIST}" == ".amzn2023" ]]; then
+      AL2023_SKIP=" perl-Sys-MemInfo ccache libudev-devel "
```
EvgeniyPatlan: ccache is available in SPAL repository

pxc/docker/install-deps
```diff
@@ -128,20 +130,42 @@ if [ -f /usr/bin/yum ]; then
...
-  if [[ ${RHVER} -eq 10 ]]; then
+  if [[ ${RHVER} -eq 10 ]] || [[ "${DIST}" == ".amzn2023" ]]; then
       PKGLIST+=" mariadb-connector-c-devel"
   else
       PKGLIST+=" mysql-devel"
   fi
 
-  # PXC specific
-  PKGLIST+=" libmicrohttpd-devel re2-devel check-devel scons libgcrypt-devel libev-devel vim-common \
-    perl-CPAN rsync python3 patchelf automake bzip2 gnutls gnutls-devel patch lsof socat stunnel pigz net-tools netcat \
```
EvgeniyPatlan: netcat is in the repo

```
bash-5.2# dnf list netcat
Last metadata expiration check: 0:10:33 ago on Thu Sep 24 09:49:11 2026.
Available Packages
netcat.x86_64                             1.238-1.1.spal2023                             amazonlinux-spal
```

EvgeniyPatlan (changes requested):

pxc/docker/install-deps
nogueiraanderson: Unchanged from master, and `dot:giga` is already wget's quietest dot mode. Leaving it as is.

pxc/docker/install-deps
nogueiraanderson: Removed in 51e8cbed.

pxc/docker/install-deps
nogueiraanderson: Agreed it should go, but `centos:7` is still in the `prepare-pxc-build-docker-ecr` image list, so removing only this block breaks that image. Better as a separate PR that drops all EL7 branches.

pxc/docker/install-deps
nogueiraanderson: Done in 51e8cbed. AL2023 now installs `spal-release`, so ccache comes from SPAL.

pxc/docker/install-deps
nogueiraanderson: Done in 51e8cbed, `netcat` now comes from SPAL on AL2023.

pxc/docker/install-deps
nogueiraanderson: The tools repo only supplies qpress to the OL8/9 and CentOS 7 images. Dropping it removes qpress test coverage there, so better as its own change.

PR #4097, opened 2026-05-19
fix(ps): build PS 8.0/9.0 source tarball in Docker instead of native Focal node

EvgeniyPatlan:
## Problem

The `Create PS source tarball` stage in both `hetzner-ps8.0-autobuild-RELEASE` and `hetzner-ps9.0-RELEASE` ran `buildStage("none", ...)` — i.e. **natively on the agent host OS** (Hetzner: `deb12-x64` / AWS: `min-focal-x64`).

With `CLOUD=AWS` this builds sources natively on the AWS spot host (Ubuntu Focal 20.04), so the build environment depends on the host rather than a pinned image. For PS 9.x the Focal toolchain/deps do not reliably satisfy the build, so `CLOUD=AWS` runs fail at the very first stage while Hetzner (Debian 12) succeeds. PS 8.0 carries the same host-OS coupling.

## Fix

Move both stages onto a generic Docker build node (`docker-x64` / `docker-32gb`) and run `get_sources` inside a pinned container, mirroring each pipeline's own `Build PS generic source deb` stage:

- **ps9.0:** `buildStage("ubuntu:jammy", "--get_sources=1")`
- **ps8.0:** `buildStage("ubuntu:focal", ...)` — FIPS and non-FIPS branches

The source-build OS is now identical on AWS and Hetzner; `CLOUD` only selects the node pool, not the build environment. All other steps (`slackNotify`, `cleanUpWS`, `installCli("deb")`, stash, `pushArtifactFolder`, `uploadTarballfromAWS`) are unchanged. `buildStage` mounts the workspace into the container so the generated `test/percona-server-*.properties` still reaches the host for the subsequent stash/upload steps.

### DIFF
--- a/ps/jenkins/percona-server-for-mysql-8.0.groovy
+++ b/ps/jenkins/percona-server-for-mysql-8.0.groovy
@@ -416,7 +416,7 @@ parameters {
     stages {
         stage('Create PS source tarball') {
             agent {
-               label params.CLOUD == 'Hetzner' ? 'deb12-x64' : 'min-focal-x64'
+                label params.CLOUD == 'Hetzner' ? 'docker-x64' : 'docker-32gb'
             }
             steps {
                 slackNotify("${SLACKNOTIFY}", "#00FF00", "[${JOB_NAME}]: starting build for ${BRANCH} - [${BUILD_URL}]")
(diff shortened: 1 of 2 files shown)


nogueiraanderson (approved):
LGTM. Aligns the `Create PS source tarball` stage with the existing `Build PS generic source {rpm,deb}` pattern in the same files (`docker-x64`/`docker-32gb` + pinned `ubuntu:focal`/`ubuntu:jammy`), so the source-build env no longer depends on the agent host OS.

Two follow-ups, non-blocking:

- `ps/jenkins/percona-server-for-mysql-8.0-arm.groovy:238-246` still pins `label 'min-focal-x64'` + `buildStage("none", "--get_sources=1")`. Same root cause; worth a sibling PR.
- Please confirm at least one green run per `CLOUD` on `hetzner-ps8.0-autobuild-RELEASE` (FIPS + non-FIPS) and `hetzner-ps9.0-RELEASE` before merging — Hetzner's path is also changing (native -> container), not only AWS.

adivinho (approved):

PR #4104, opened 2026-05-20
chore(scripts): move install-master-observability.sh to percona-cd-platform

nogueiraanderson:
## Summary

Move the canonical \`install-master-observability.sh\` to the platform repo where the gateway, IAM grant (\`AlloyGatewayBearerRead\`), and Secrets Manager bearer it depends on already live: [percona/percona-cd-platform@6f5346e](https://github.com/percona/percona-cd-platform/blob/6f5346eadc4bfa44e6d86d73ceb62ce7efbb7b79/scripts/install-master-observability.sh). Both repos are AGPL-3.0; no licence boundary.

This PR turns the file in jenkins-pipelines into a thin shim:

\`\`\`bash
exec bash -c "$(curl -fsSL --retry 5 \"$CANONICAL_URL\")"
\`\`\`

The shim is SHA-pinned to the same commit the platform's Terraform user-data points at, so masters fetching either URL run identical content.

## Compatibility

- Existing CFN-managed Jenkins masters pin to a specific SHA of this file and fetch it from GitHub raw. Those URLs serve immutable content by SHA; current launches and re-launches are unaffected and continue running the original 128-line script.
- Anyone fetching at master HEAD now gets the shim, which curls the canonical version and pipes to bash. \`JENKINS_HOST\` and any other caller env vars pass through.
- The platform's TF user-data (ps3 and pxb today) already points at the new URL.

## When this file can be deleted

When the last CFN-managed master migrates to Terraform in percona-cd-platform.

### DIFF
--- a/scripts/install-master-observability.sh
+++ b/scripts/install-master-observability.sh
@@ -1,128 +1,24 @@
 #!/bin/bash
 # install-master-observability.sh
 #
-# PS-10997 / ADR 0013: master-side push pipeline.
-# Installs amazon-ssm-agent (fleet ops fan-out) and Grafana Alloy
-# (scrapes /hetzner-prometheus + tails Jenkins log; pushes to the
-# in-cluster alloy-gateway ALB). Bearer token is fetched from AWS
-# Secrets Manager at every Alloy restart via systemd ExecStartPre.
+# Back-compat shim. The canonical script now lives in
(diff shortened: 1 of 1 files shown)


nogueiraanderson:
Closing per maintainer request. Will revisit if the back-compat shim becomes needed (e.g. when more masters migrate off CFN and start fetching from this file's master HEAD).

PR #4127, opened 2026-05-28
Pbm 1710 pbm docker add SBOM

vorsel:
New job hetzner-pbm-docker-multiarch builds amd64 and arm64 PBM images in
parallel, generates per-arch CycloneDX 1.6 SBOMs from the locally-built
images (BEFORE push, so a broken SBOM fails the whole release), pushes
per-arch tags, assembles a multi-arch manifest, then attaches SBOMs as
OCI 1.1 referrer artifacts via 'oras attach' for all three targets
(PerconaLab, DockerHub, AWS_ECR).


nogueiraanderson (approved):
LGTM. Sibling parity with PCSM 4128 is clean.

olexandr-havryliak (approved):

keithquinnpercona (approved):

PR #4128, opened 2026-05-28
Pcsm 282 pcsm docker add SBOM

vorsel:
Builds linux/amd64 and linux/arm64 PCSM container images in parallel,
generates per-arch CycloneDX 1.6 SBOMs from the locally-built images,
pushes per-arch tags, assembles a multi-arch manifest via
docker buildx imagetools, then attaches the SBOMs as OCI 1.1 referrer
artifacts via oras attach.

Mirrors the PBM Docker multi-arch + SBOM pipeline (PBM-1710); the only
intentional difference is that PCSM does not publish to AWS ECR, so
TARGET_REPO is limited to PerconaLab and DockerHub only.

### DIFF
--- a/pcsm/pcsm-docker-multiarch.groovy
+++ b/pcsm/pcsm-docker-multiarch.groovy
@@ -1,6 +1,4 @@
-// PCSM Docker multi-arch + SBOM pipeline.
-//
-// Builds linux/amd64 and linux/arm64 PCSM container images in parallel,
+// Build PCSM container images in parallel,
 // generates per-arch CycloneDX 1.6 SBOMs from the locally-built images,
 // pushes per-arch tags, assembles a multi-arch manifest, then attaches the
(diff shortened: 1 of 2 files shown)


nogueiraanderson:
LGTM. Mirrors the PBM sibling pair faithfully. Packaging correctly fixes the silent `get_tar` TARBALL clobber and moves `build_tarball` ahead of source builds so the new `exit 1` in the SBOM block surfaces real failures. One small nit on 4128: the syft cataloger list includes `dpkg-db-cataloger` but the PCSM image is UBI9 rpm-only, so dpkg-db never matches anything. Harmless, but worth trimming for consistency with the documented rpm-db for UBI9 scope.

nogueiraanderson (approved):
LGTM. See inline comment for the optional dpkg-db-cataloger nit.

sandraromanchenko (approved):

sandraromanchenko (approved):

PR #3945, opened 2026-04-07
PKG-1306: make chown -R non-fatal in start_jenkins() to prevent bootstrap abort

nogueiraanderson:
## Summary

- Add `|| true` to `chown -R jenkins:jenkins` in `start_jenkins()` across all 12 IaC templates

Broken symlinks in JENKINS_HOME (e.g. `queue.xml`) cause `chown` to exit non-zero. With `set -o errexit`, this kills the entire UserData script before Jenkins, OpenResty, or certbot can start.

## Context

pmm.cd.percona.com went down on 2026-04-07 after a spot instance replacement. The new instance had UserData partially execute but abort at the `chown -R` step in `start_jenkins()`:

```
chown: changing ownership of '/mnt/pmm.cd.percona.com/queue.xml': No such file or directory
```

cloud-init logged `Failed to run module scripts-user` and the instance came up as a bare Amazon Linux box with no Jenkins, no reverse proxy, and no SSL.

Jira: [PKG-1306](https://perconadev.atlassian.net/browse/PKG-1306)

[PKG-1306]: https://perconadev.atlassian.net/browse/PKG-1306?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #4137, opened 2026-05-29
PKG-1335: harden Jenkins master bootstrap against transient dnf failures

nogueiraanderson:
## Bug

- A Spot Fleet replacement of a Jenkins master ran its cloud-init UserData, which aborted under `set -o errexit` when `yum -y install java-17-amazon-corretto` hit a transient dnf package-cache miss. Jenkins and OpenResty never started and `pmm.cd` was unreachable for ~2.5h.

## Fix

- Add a `retry()` helper (retry up to 5 times, fail fast after) and route every `yum -y install` and `yum -y update --security` through it.
- Gate the recursive `chown` of `JENKINS_HOME` on a `FRESH_DATA_VOLUME` flag, set when `mkfs.xfs` succeeds on a blank volume. Reattached homes skip the ~13 min walk; fresh volumes still get the full chown.
- Applied to the five CFN masters: `cloud`, `pg`, `pmm`, `psmdb`, `rel`. The Terraform-managed masters (`ps3`, `ps80`, `ps57`, `pxb`, `pxc`) are handled in percona-cd-platform, not here.

## Tickets

- [PKG-1335](https://perconadev.atlassian.net/browse/PKG-1335): this
- [PKG-1306](https://perconadev.atlassian.net/browse/PKG-1306): same bootstrap-abort class ([PR 3945](https://github.com/Percona-Lab/jenkins-pipelines/pull/3945))

[PKG-1335]: https://perconadev.atlassian.net/browse/PKG-1335?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[PKG-1306]: https://perconadev.atlassian.net/browse/PKG-1306?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/IaC/pmm.cd/JenkinsStack.yml
+++ b/IaC/pmm.cd/JenkinsStack.yml
@@ -622,13 +622,26 @@ Resources:
                            Key=iit-billing-tag,Value=$JENKINS_SHORT
             }
 
+            retry() {
+                local attempt=1
+                until "$@"; do
+                    if [ $attempt -ge 5 ]; then
+                        echo "command failed after $attempt attempts: $*" >&2
+                        return 1
+                    fi
+                    echo "command failed (attempt $attempt), retrying: $*"
+                    attempt=$((attempt + 1))
+                    sleep 5
+                done
+            }
+
             install_software() {
                 until yum makecache; do
                     sleep 1
                     echo try again
                 done
 
-                yum -y install yum-utils wget cronie rsync bc
+                retry yum -y install yum-utils wget cronie rsync bc
                 yum-config-manager --add-repo "https://openresty.org/package/amazon/2023/x86_64/"
                 rpm --import https://openresty.org/package/pubkey.gpg
                 wget -O /etc/yum.repos.d/jenkins.repo https://pkg.jenkins.io/redhat-stable-legacy/jenkins.repo
@@ -636,9 +649,9 @@ Resources:
                 grep -q 'repo_gpgcheck=0' /etc/yum.repos.d/jenkins.repo
                 rpm --import https://pkg.jenkins.io/redhat-stable-legacy/jenkins.io-2023.key
                 rpm --import https://pkg.jenkins.io/redhat-stable/jenkins.io-2026.key
-                yum -y install java-17-amazon-corretto
-                yum -y update --security
-                yum -y install jenkins-2.528.3 certbot git dnf-automatic aws-cli xfsprogs openresty
+                retry yum -y install java-17-amazon-corretto
+                retry yum -y update --security
+                retry yum -y install jenkins-2.528.3 certbot git dnf-automatic aws-cli xfsprogs openresty
 
                 sed -i 's/upgrade_type = default/upgrade_type = security/' /etc/dnf/automatic.conf
                 sed -i 's/apply_updates = no/apply_updates = yes/'         /etc/dnf/automatic.conf
@@ -675,7 +688,11 @@ Resources:
                     [ -e /dev/xvdj ] && break || sleep 1
                 done
 
-                mkfs.xfs -L DATA /dev/xvdj || :
+                if mkfs.xfs -L DATA /dev/xvdj; then
+                    FRESH_DATA_VOLUME=1
+                else
+                    FRESH_DATA_VOLUME=0
+                fi
                 echo "/dev/xvdj /mnt xfs defaults,noatime,nofail 0 0" | tee -a /etc/fstab
                 mount /mnt
             }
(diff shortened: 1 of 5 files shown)


IaC/pmm.cd/JenkinsStack.yml
```diff
@@ -622,23 +622,38 @@ Resources:
...
+            yum_retry() {
+                local _yum_try=1
+                until "$@"; do
+                    if [ $_yum_try -ge 5 ]; then
+                        echo "command failed after $_yum_try attempts: $*" >&2
+                        return 1
+                    fi
+                    echo "command failed (attempt $_yum_try), cleaning and retrying: $*"
+                    yum clean all || true
+                    yum makecache || true
```
ademidoff:
```suggestion
```

We have already run `makecache` e.g. line 641. While retrying a package install is indeed useful, I don't believe multiple `clean all` and `makecache` solve anything. I think it's redundant and it slows things down.

ademidoff (changes requested):

IaC/pmm.cd/JenkinsStack.yml
```diff
@@ -699,7 +718,16 @@ Resources:
                 install -o jenkins -g jenkins -d /mnt/$JENKINS_HOST
                 install -o jenkins -g jenkins -d /mnt/$JENKINS_HOST/init.groovy.d
                 install -o jenkins -g jenkins -d /var/log/jenkins
-                chown -hR jenkins:jenkins /mnt/$JENKINS_HOST /var/log/jenkins
+                chown -h jenkins:jenkins /mnt/$JENKINS_HOST
```
ademidoff:
```suggestion
```

It's harmless, but redundant since the `install -o jenkins...` handles it.

IaC/pmm.cd/JenkinsStack.yml
nogueiraanderson: Good call. Went pure retry, dropped both `clean all` and `makecache`. The failure is a concurrent dnf (the boot `dnf-makecache.timer` / SSM auto-update) sweeping the shared package cache under `keepcache=false`; on retry dnf just re-downloads the missing payload, so no cache clearing is needed, the retry only has to outlast the racer. Thanks for pushing on it.

IaC/pmm.cd/JenkinsStack.yml
nogueiraanderson: Good catch, removed it, and the `/var/log/jenkins` recursive chown too for the same reason since the `install -o jenkins -g jenkins -d` lines already set ownership. The only chown left now is the recursive one, gated to run solely on a genuinely fresh data volume.

IaC/pmm.cd/JenkinsStack.yml
```diff
@@ -622,23 +622,36 @@ Resources:
...
 
+            yum_retry() {
+                local _yum_try=1
+                until "$@"; do
+                    if [ $_yum_try -ge 5 ]; then
+                        echo "command failed after $_yum_try attempts: $*" >&2
+                        return 1
+                    fi
+                    echo "command failed (attempt $_yum_try), retrying: $*"
+                    _yum_try=$((_yum_try + 1))
```
ademidoff:
```suggestion
                local attempt=1
                until "$@"; do
                    if [ $attempt -ge 5 ]; then
                        echo "command failed after $attempt attempts: $*" >&2
                        return 1
                    fi
                    echo "command failed (attempt $attempt), retrying: $*"
                    ((attempt++))
```

I'm a bit surprised by the naming of the `_yum_try` variable. 

1. In the function scope `local` does not leak the variables outside, so it's unclear why we need to use an underscore in the name, which basically translates to "local local yum_try" to the human. All it does is increase the cognitive load to the reader.
2. The "yum_" prefix add a specific sope to this variable. I wonder if something as generic as "attempt" cannot be a much better choice. Simpler and more understandable as it removes the unnecessary scope prefix.

nogueiraanderson: Good naming call, applied it: the counter is just `attempt` now (dropped both the `_` and the `yum_` prefix), across all six CFN files plus the pxb userdata for consistency.

One note on the suggestion itself: I kept the increment as `attempt=$((attempt + 1))` rather than `((attempt++))`. Under `set -o errexit` the `((...))` arithmetic command returns exit status 1 when the expression evaluates to 0, and post-increment evaluates to the pre-increment value. 

It's safe here only because the counter starts at 1; if it ever started at 0 or got refactored, the first `((attempt++))` would abort the bootstrap, which is the exact failure class this PR is hardening against. 

The assignment form always returns 0 and matches the existing `RETRY=$(($RETRY-1))` lower in the file. Thanks for pushing on this.

ademidoff (approved):

IaC/pmm.cd/JenkinsStack.yml
ademidoff: 1. This counter always starts with 1
2. The reliability fix is easy - just put a colon before `: ((attempt++))`

EvgeniyPatlan (approved):

PR #4158, opened 2026-06-03
PMM-15119 Update Trivy to 0.71.0 and add htmlTpl option

talhabinrizwan:
* Update Trivy to 0.71.0 and add htmlTpl option
* Migrate pmm3-image-scanning to installTrivy shared library

### DIFF
--- a/vars/installTrivy.groovy
+++ b/vars/installTrivy.groovy
@@ -8,14 +8,16 @@
 //   installTrivy(method: 'binary')          // force GitHub tar.gz method
 //   installTrivy(method: 'apt')             // force APT method
 //   installTrivy(method: 'binary', junitTpl: true)   // also download junit.tpl
+//   installTrivy(htmlTpl: true)                      // also download html.tpl
 
 def call(Map args = [:]) {
-    def TRIVY_VERSION    = "0.70.0"
-    def CHECKSUM_AMD64   = "8b4376d5d6befe5c24d503f10ff136d9e0c49f9127a4279fd110b727929a5aa9"
-    def CHECKSUM_ARM64   = "2f6bb988b553a1bbac6bdd1ce890f5e412439564e17522b88a4541b4f364fc8d"
+    def TRIVY_VERSION    = "0.71.0"
+    def CHECKSUM_AMD64   = "30a3d22b23f88c233f1658f562fb477cae3b3e8b4761109d515b7698daf85814"
+    def CHECKSUM_ARM64   = "2561be394a3199c911f82fced606cbc05e1cb23eb6ce1da6935540adb76f4252"
 
     def method    = args.get('method', 'auto')
     def junitTpl  = args.get('junitTpl', false)
+    def htmlTpl   = args.get('htmlTpl', false)
 
     if (method == 'auto') {
         // Detect: use APT on Debian/Ubuntu, binary elsewhere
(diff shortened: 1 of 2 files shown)


vars/installTrivy.groovy
```diff
@@ -71,4 +73,12 @@ def call(Map args = [:]) {
             fi
         """
     }
+
+    if (htmlTpl) {
+        sh """
+            if [ ! -f html.tpl ]; then
+                wget -q https://raw.githubusercontent.com/aquasecurity/trivy/v${TRIVY_VERSION}/contrib/html.tpl
+            fi
```
EvgeniyPatlan:
```suggestion
TRIVY_VERSION="${TRIVY_VERSION:-0.71.0}"

declare -A TPL_SHA256=(
    ["0.71.0"]="d36b637cc77bc1a49b1f9a2358173756cd27a5d3385fab6365c0bab09abe5f0a"
)

expected_sha="${TPL_SHA256[$TRIVY_VERSION]:?No pinned html.tpl checksum for Trivy ${TRIVY_VERSION}. Verify and add one before building}"

if [ ! -f html.tpl ] || ! echo "${expected_sha}  html.tpl" | sha256sum -c - >/dev/null 2>&1; then
    tmp="$(mktemp)"
    wget -q -O "$tmp" \
        "https://raw.githubusercontent.com/aquasecurity/trivy/v${TRIVY_VERSION}/contrib/html.tpl" \
        || { echo "html.tpl download failed" >&2; rm -f "$tmp"; exit 1; }
    echo "${expected_sha}  ${tmp}" | sha256sum -c - \
        || { echo "html.tpl checksum mismatch for v${TRIVY_VERSION} — refusing to use untrusted template" >&2; rm -f "$tmp"; exit 1; }
    mv "$tmp" html.tpl
fi
```

EvgeniyPatlan (changes requested):

vars/installTrivy.groovy
```diff
@@ -71,4 +73,26 @@ def call(Map args = [:]) {
...
+        sh """
+TRIVY_VERSION="${TRIVY_VERSION:-0.71.0}"
+
+declare -A TPL_SHA256=(
+    ["0.71.0"]="d36b637cc77bc1a49b1f9a2358173756cd27a5d3385fab6365c0bab09abe5f0a"
+)
+
+expected_sha="${TPL_SHA256[$TRIVY_VERSION]:?No pinned html.tpl checksum for Trivy ${TRIVY_VERSION}. Verify and add one before building}"
+
+if [ ! -f html.tpl ] || ! echo "${expected_sha}  html.tpl" | sha256sum -c - >/dev/null 2>&1; then
```
ademidoff: First time I see `echo` being prefixed with "!". Is it intentional?

ademidoff (approved):

vars/installTrivy.groovy
ademidoff: Frankly, it's not readable for a human at all. I'll end up hating that AI slop :)))

EvgeniyPatlan (approved):

vars/installTrivy.groovy
```diff
@@ -71,4 +73,26 @@ def call(Map args = [:]) {
             fi
         """
     }
+
+    if (htmlTpl) {
+        withEnv(["TRIVY_VERSION=${TRIVY_VERSION}"]) {
```
ademidoff: @talhabinrizwan Given that `TRIVY_VERSION` is a variable within the global scope, why do we need to pass it here like that?

To me, all it does it reduces the readability and complicates maintenance.

talhabinrizwan: `TRIVY_VERSION` is a Groovy local, so it isn't visible to bash unless we either 
(a) interpolate it into the script via sh """...""", which forces us to escape every bash $ as \\$, or 
(b) push it into the child env with withEnv and keep sh '''...''' so the bash reads cleanly. 
Went with (b) for readability.

ademidoff (approved):

PR #4166, opened 2026-06-04
PSMDB-945 psmdb.cd: add mongot build pipelie

vorsel:


sandraromanchenko (approved):

olexandr-havryliak (approved):

nogueiraanderson:
Reviewed against the psmdb-8.0 build sibling (`get-psmdb-branches-8.0.groovy`, `percona-server-for-mongodb-8.0.groovy`) and the shared `vars/` library. Structurally this tracks the psmdb family closely. Grouped by whether it is specific to this pipeline or inherited from the siblings:

**ps4m-specific**
- `RPM_RELEASE` and `DEB_RELEASE` are declared (lines 62-69) and forwarded by the poller, but the builder invocation passes only `--version` (line 29), so they never reach `mongot_builder.sh`. The psmdb sibling passes `--psm_release`. Either wire them in or drop them.
- The x86_64 and aarch64 bundles push to the same `bazel_tarball/` and `tarball/` prefixes (lines 133-134, 148-149), and every RPM/DEB stage pops the whole `bazel_tarball/` (line 165 onward). Shared `tarball/` has precedent in the psmdb sibling, but `bazel_tarball/` is new here, so each package stage pulls both arches' bundles. If any emitted filename is not arch-qualified the two collide; if they are, it is wasted transfer plus the wrong-arch bundle landing in each stage. Arch-scoped prefixes would remove the ambiguity.
- No smoke or functional check is wired. The psmdb sibling hands off to a downstream test multijob; mongot has none yet (PSMDB-2032 tracks package/tarball tests). PSMDB-1945's DoD lists a `--version` smoke test, so worth confirming where it lands before this is considered done.

**Worth hardening (inherited from the psmdb siblings, so equally a family-wide follow-up)**
- Branch discovery fails open: the `git ls-remote ... | tail -1` block in `get-ps4m-branches.groovy` runs without `set -euo pipefail`, so an auth/network error yields an empty result and `START_NEW_BUILD=NO` (lines 61-65) rather than a hard failure. The psmdb-8.0 sibling has the same gap. Adding `set -euo pipefail` to the discovery `sh` would make a fetch failure loud.
- The new-commit marker is written to S3 (line 84) before the downstream build is triggered (line 100), so a failed build is not retried on the next poll. Same shape as the sibling. If poll-driven retry is wanted, persist the marker only after the build succeeds.
- The `mongot_builder.sh` script is fetched from a mutable branch over `wget` and run under `docker run -u root` with no checksum (lines 20, 25), matching the sibling. If `percona/percona-mongot` is private, `git ls-remote` / `wget` / clone also need a credential path. Worth confirming the repo is public by release time or wiring a token.
- The inner Docker shell sets `xtrace` only (lines 8, 26); `bash -euo pipefail` would stop a partial install/build step from continuing silently. Separately, `vars/pushArtifactFolder.groovy` ends its `aws s3 cp` with `|| :` inside `retry(15)` (line 10), so a failed upload is swallowed and never actually retried.

**Ticket**
- PSMDB-1945 scopes "a tarball for x64 platform now" and says clone `mongodb/mongot`, but this builds amd64 + aarch64 plus RPM/DEB and defaults `GIT_REPO` to `percona/percona-mongot` (line 51). The wider arch/packaging scope likely matches the epic, and a Percona fork is reasonable for carrying patches, but the ticket text and DoD (smoke test, Notion build-flag docs) should be updated to match what shipped.

ps4m/jenkins/percona-search-for-mongodb.groovy
```diff
@@ -0,0 +1,384 @@
...
+                    }
+                    steps {
+                        cleanUpWS()
+                        unstash 'mongot-properties'
+                        popArtifactFolder(params.CLOUD, "source_tarball/", AWS_STASH_PATH)
+                        script {
+                            buildStage("oraclelinux:8", "--build_mongot=1 --build_variant=linux-x64")
+                        }
+                        pushArtifactFolder(params.CLOUD, "bazel_tarball/", AWS_STASH_PATH)
+                        pushArtifactFolder(params.CLOUD, "tarball/", AWS_STASH_PATH)
```
nogueiraanderson: Both arch bundle stages push to the same arch-less S3 prefixes, and every per-OS package stage pops the whole bundle with no arch selection. `Bundle x86_64` pushes `bazel_tarball/` and `tarball/` here (lines 129-130); `Bundle aarch64` pushes the [same prefixes](https://github.com/Percona-Lab/jenkins-pipelines/blob/c3cadc1a9628d5faeea48eda221026d7bd2ee970/ps4m/jenkins/percona-search-for-mongodb.groovy#L144-L145) under one `AWS_STASH_PATH`. Each RPM/DEB stage then pops the [whole `bazel_tarball/`](https://github.com/Percona-Lab/jenkins-pipelines/blob/c3cadc1a9628d5faeea48eda221026d7bd2ee970/ps4m/jenkins/percona-search-for-mongodb.groovy#L161) and runs `--build_rpm`/`--build_deb` with no arch flag. If the builder's bundle filenames are not arch-qualified, a stage can package the wrong-arch binary, which is then signed and synced to the public repo. The mongosh and psmdb pipelines avoid this by popping arch-agnostic source and building per-arch in-stage.

Illustrative fix (not a one-click suggestion, since it spans the two pushes plus the ~11 pop sites): arch-scope the intermediate prefixes so the two arches never share a key, and have each stage pop only its own arch into a plain working directory.

```groovy
// Bundle x86_64
pushArtifactFolder(params.CLOUD, "bazel_tarball/x86_64/", AWS_STASH_PATH)
// Bundle aarch64
pushArtifactFolder(params.CLOUD, "bazel_tarball/aarch64/", AWS_STASH_PATH)
// each x86_64 package stage (and bazel_tarball/aarch64/ for the arm64 stages):
popArtifactFolder(params.CLOUD, "bazel_tarball/x86_64/", AWS_STASH_PATH)
```

Equivalent alternative: fold packaging into each bundle stage so the arch artifact never leaves its producing stage, matching mongosh/psmdb.

nogueiraanderson (approved):
Approving to land the pipeline. The points from my earlier review (RPM_RELEASE/DEB_RELEASE not reaching mongot_builder.sh, the shared bazel_tarball/ prefix across both arches, no smoke test wired yet, and the fail-open branch discovery in get-ps4m-branches.groovy) are non-blocking follow-ups rather than merge blockers.

nogueiraanderson (approved):
Skimmed it: the mongot build pipeline (get-ps4m-branches + percona-search-for-mongodb) is clean, additive, and follows the usual JJB pattern. LGTM.

ps4m/jenkins/percona-search-for-mongodb.groovy
vorsel: Reworked

PR #4167, opened 2026-06-04
PSMDB-1945 ps4m: add Percona Search for MongoDB (mongot) docker build pipeline

vorsel:
Multi-arch (amd64 + arm64) container build for mongot.


sandraromanchenko (approved):

olexandr-havryliak (approved):

nogueiraanderson:
Reviewed against `pbm/pbm-docker-multiarch.groovy` and `pcsm/pcsm-docker-multiarch.groovy`. Native per-arch build plus `buildx imagetools` manifest assembly is the right shape, and the image name / Dockerfile paths / registry are consistent with percona/percona-docker#1349. A few things:

**Should-fix**
- Cleanup only exists in the top-level `post`, which runs on the manifest agent, not on the `docker-x64` / `docker-aarch64` build workers. PBM/PCSM clean each parallel worker in a stage-local `post { always { docker rmi ...; deleteDir() } }`. Without it the multi-GB images accumulate on the reused Hetzner nodes.
- `LATEST` defaults to `yes` and the `percona` prod namespace is selectable with no gate tying it to a release channel, so a preview build can publish `percona/percona-server-mongodb-mongot:latest` by a parameter mistake. PCSM has the same shape, so it is a shared convention rather than new here, but a fail-closed guard before login/push (here or fleet-wide) would prevent it.
- Params (`MONGOT_VERSION`, `DOCKER_BRANCH`, `DOCKER_REPO`, `PSMDB_REPO`) are interpolated into `sh`, and `MONGOT_VERSION` into the Docker Hub credential block. This matches the existing pbm/pcsm pattern, so it is a fleet-wide convention rather than new here, but quoting/validating the values would be safer.

**Nits**
- The AWS amd64 stage uses `docker-64gb`, which is registered only on psmdb.cd. The siblings' amd64 uses the more widely-registered `docker-32gb`, though their arm64 uses `docker-64gb-aarch64`, which is also psmdb-only, so all three are effectively psmdb-pinned for `CLOUD=AWS`. The default is `Hetzner` (resolves everywhere) and only the Hetzner job is defined, so this only bites if someone runs it with `CLOUD=AWS` off psmdb. A one-line comment pinning the AWS path to psmdb.cd would make that explicit.
- No SBOM/Trivy step, which both the pbm and pcsm multi-arch pipelines include. Worth matching if this is release-capable, or keeping explicitly preview-only.
- Reports to `#releases-ci`; the pbm/pcsm multi-arch docker convention is `#mongodb_autofeed`.
- `VCS_REF` / `BUILD_DATE` build-args are not passed, so the `org.opencontainers.image.revision` / `.created` labels the Dockerfile declares stay `unknown`.

**Ticket**
- Like #1349, the container build sits outside PSMDB-1945's stated scope; a dedicated container ticket would give it a clear home and CI acceptance criteria.

nogueiraanderson (approved):
LGTM. Per-arch build and `buildx imagetools` manifest assembly look correct, and the tag gating is sound: `:latest` is fail-closed to the `release` channel and major-only tags are held while mongot is on the 0.x series. The version/channel sed targets line up with the Dockerfile ENV names in percona-docker#1349.

nogueiraanderson (approved):
Skimmed it: the multiarch docker build pipeline is clean and additive, consistent with the build pipeline in #4166. LGTM.

PR #4171, opened 2026-06-05
PS-11257: reap leaked molecule package-testing EC2 instances

nogueiraanderson:
**Bug**
- Aborted package-testing molecule builds skip `molecule destroy`, leaking EC2 instances indefinitely; their category `iit-billing-tag` (`*_package_testing`) exempts them from the cleanup Lambda forever.

**Fix**
- `timeout(time: 6, unit: 'HOURS')` on the 4 molecule pipelines so an abort still reaches the `post { always }` destroy.
- The Lambda-side backstop (7h age-bound on `.*_package_testing$` billing tags) now lives in the Terraform reaper: [percona-cd-platform#123](https://github.com/percona/percona-cd-platform/pull/123). This PR is groovy-only and no longer touches `IaC/LambdaEC2Cleanup.yml` (deleted by [#4183](https://github.com/Percona-Lab/jenkins-pipelines/pull/4183)).

**Tickets**
- [PS-11257](https://perconadev.atlassian.net/browse/PS-11257)

[PS-11257]: https://perconadev.atlassian.net/browse/PS-11257?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #4178, opened 2026-06-08
PS-11262: fix volume-cleanup Lambda to reap untagged volumes

nogueiraanderson:
**Bug**
- `LambdaVolumeCleanup` filters volumes by `tag:Name=*`, so it only evaluates Name-tagged volumes and silently skips untagged unattached ones. Two untagged 30 TB gp3 volumes ran for days (a dev-account cost anomaly) and untagged orphans had accumulated for months.

**Fix**
- Evaluate every `available` volume regardless of tags. Preserve the opt-out (`PerconaKeep` tag, `"do not remove"` in `Name`).
- Guard `volume.tags == None`, add a `MIN_AGE_HOURS` (24h) floor so freshly-created volumes are not reaped mid-workflow, and move `python3.8` to `python3.12`.
- Validated on an isolated region-scoped clone. Dry run flagged the target, the real run deleted it.

**Note**
- The 8-region scan list stays hardcoded. Orphans in any other enabled region are never reaped (all empty today). Left as a follow-up.

**Tickets**
- PS-11262. Sibling to #4171 (same leak class, `LambdaEC2Cleanup`).

### DIFF
--- a/IaC/LambdaVolumeCleanup.yml
+++ b/IaC/LambdaVolumeCleanup.yml
@@ -46,33 +46,60 @@ Resources:
     Properties: 
       FunctionName: LambdaVolumeCleanup
       Description: Cleans up unused volumes
-      Runtime: python3.8
+      Runtime: python3.12
       Handler: index.lambda_handler
       MemorySize : 128
       Timeout: 240
+      Environment:
+        Variables:
+          MIN_AGE_HOURS: "24"
       Code: 
         ZipFile: |
+          import os
           import boto3
+          from datetime import datetime, timezone, timedelta
+
+          REGIONS = ["us-east-1", "us-east-2", "us-west-1", "us-west-2",
+                     "eu-central-1", "eu-west-1", "eu-west-2", "eu-west-3"]
+          DRY_RUN = os.environ.get("DRY_RUN", "false").lower() == "true"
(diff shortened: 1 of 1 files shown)


nogueiraanderson:
#### Closing: superseded by the Terraform port

The volume-cleanup Lambda moved out of CloudFormation entirely. It now lives in [Percona/percona-cd-platform](https://github.com/percona/percona-cd-platform) as OpenTofu ([#113](https://github.com/percona/percona-cd-platform/pull/113), tests [#117](https://github.com/percona/percona-cd-platform/pull/117), docs + [ADR 0030](https://github.com/percona/percona-cd-platform/blob/main/docs/adr/0030-account-cleanup-reapers-in-terraform.md) [#119](https://github.com/percona/percona-cd-platform/pull/119)).

Everything this PR fixed shipped there, locked by a moto regression suite: untagged and tag-less volumes are evaluated (the `tag:Name=*` filter is gone), `MIN_AGE_HOURS` floor, `DRY_RUN` gate, and the `PerconaKeep` / `do not remove` opt-outs.

The legacy `LambdaVolumesCleanup` stack was decommissioned (its shared schedule rule retained for the two PMM cleanup functions). The new reaper is deployed and validated in dry-run. Deleting the now-dead `IaC/LambdaVolumeCleanup.yml` + `IaC/LambdaEC2Cleanup.yml` templates is a follow-up to this repo.

PR #4202, opened 2026-06-15
PMM-15127: Resolve post-release test versions from Docker Hub

travagliad:
## Summary
- Post-release pipeline resolves latest and previous PMM versions from released \percona/pmm-server\ tags on Docker Hub instead of the static \pmmVersion('v3')\ list.
- Avoids running against a version that was added to the map before GA (e.g. 3.8.1 in the list while Hub still has 3.8.0).

### DIFF
--- a/pmm/v3/pmm3-post-release-tests.groovy
+++ b/pmm/v3/pmm3-post-release-tests.groovy
@@ -8,10 +8,6 @@ library changelog: false, identifier: 'v3lib@master', retriever: modernSCM(
   libraryPath: 'pmm/v3/'
 )
 
-def versionsList = pmmVersion('v3')
-def latestVersion = versionsList.last()
-def prevVersion = versionsList[-2]
-
 pipeline {
     agent {
         label 'agent-amd64'
@@ -35,6 +31,16 @@ pipeline {
             steps {
                 script {
                     env.ADMIN_PASSWORD = params.ADMIN_PASSWORD
+                    def tags = sh(
+                        script: '''wget -q "https://registry.hub.docker.com/v2/repositories/percona/pmm-server/tags?page_size=100" -O - \
+                            | jq -r '.results[].name' | grep -E '^[0-9]+\\.[0-9]+\\.[0-9]+$' | sort -V''',
+                        returnStdout: true
+                    ).trim().split('\n').findAll { it }
+                    if (tags.size() < 2) {
+                        error('Need at least two released percona/pmm-server tags on Docker Hub')
+                    }
+                    def prevVersion = tags[-2]
+                    def latestVersion = tags[-1]
                     env.PMM_SERVER_LATEST = latestVersion
                     env.DOCKER_TAG = "percona/pmm-server:${prevVersion}"
                     env.DOCKER_TAG_UPGRADE = "percona/pmm-server:${latestVersion}"
--- a/pmm/v3/pmm3-post-release-tests.groovy
+++ b/pmm/v3/pmm3-post-release-tests.groovy
@@ -36,9 +36,6 @@ pipeline {
                             | jq -r '.results[].name' | grep -E '^[0-9]+\\.[0-9]+\\.[0-9]+$' | sort -V''',
                         returnStdout: true
                     ).trim().split('\n').findAll { it }
-                    if (tags.size() < 2) {
-                        error('Need at least two released percona/pmm-server tags on Docker Hub')
-                    }
                     def prevVersion = tags[-2]
                     def latestVersion = tags[-1]
                     env.PMM_SERVER_LATEST = latestVersion
--- a/pmm/v3/pmm3-post-release-tests.groovy
+++ b/pmm/v3/pmm3-post-release-tests.groovy
@@ -33,12 +33,13 @@ pipeline {
                     env.ADMIN_PASSWORD = params.ADMIN_PASSWORD
                     def tags = sh(
                         script: '''set -e
-curl -sf "https://registry.hub.docker.com/v2/repositories/percona/pmm-server/tags?page_size=5" \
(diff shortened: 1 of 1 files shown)


pmm/v3/pmm3-post-release-tests.groovy
```diff
@@ -35,6 +31,13 @@ pipeline {
             steps {
                 script {
                     env.ADMIN_PASSWORD = params.ADMIN_PASSWORD
+                    def tags = sh(
+                        script: '''wget -q "https://registry.hub.docker.com/v2/repositories/percona/pmm-server/tags?page_size=100" -O - \
```
ademidoff: Do we need to pull 100 versions? I understand we can easily reduce that to just 5 (if not less 😄 ).

pmm/v3/pmm3-post-release-tests.groovy
```diff
@@ -8,10 +8,6 @@ library changelog: false, identifier: 'v3lib@master', retriever: modernSCM(
   libraryPath: 'pmm/v3/'
 )
 
-def versionsList = pmmVersion('v3')
```
ademidoff, line 11: Don't we control what `pmmVersion` returns? If we do, why remove it?

pmm/v3/pmm3-post-release-tests.groovy
ademidoff: This query relies on the `happy` path. What if it fails? Looks very fragile.

ademidoff: Every other network call in this file uses `curl`, this one deviates from the de-facto standard and uses `wget`, which is not guaranteed to exist.

I suggest we stop using `wget` in the pipelines, and make `curl` the standard. 

Apart from that, `curl` is simpler - `curl -s` is all we need.

pmm/v3/pmm3-post-release-tests.groovy
```diff
@@ -35,6 +31,13 @@ pipeline {
             steps {
                 script {
                     env.ADMIN_PASSWORD = params.ADMIN_PASSWORD
+                    def tags = sh(
+                        script: '''wget -q "https://registry.hub.docker.com/v2/repositories/percona/pmm-server/tags?page_size=100" -O - \
+                            | jq -r '.results[].name' | grep -E '^[0-9]+\\.[0-9]+\\.[0-9]+$' | sort -V''',
```
ademidoff: I realized the regexp silently matches all major versions, from 1 to 3. Do we need to worry about v1 and v2?

travagliad: As this is meant for post-release, I don't think we need since we're sorting results. But this is a cheap change, fixed 3 on the regex.

pmm/v3/pmm3-post-release-tests.groovy
travagliad: Changed to curl -sf so we also have explicit message for connection errors

pmm/v3/pmm3-post-release-tests.groovy
travagliad: We do, but it's changed before GA, so our post-release pipeline isn't able to run after RC.

pmm/v3/pmm3-post-release-tests.groovy
travagliad: Added conditional to check if we have at least 2 versions.

Pulling 5 versions.

travagliad: Less than 5 may not work as we have other tags that are launched (3, 3.8 and so on)

pmm/v3/pmm3-post-release-tests.groovy
```diff
@@ -35,6 +31,17 @@ pipeline {
             steps {
                 script {
                     env.ADMIN_PASSWORD = params.ADMIN_PASSWORD
+                    def tags = sh(
+                        script: '''set -e
+curl -sf "https://registry.hub.docker.com/v2/repositories/percona/pmm-server/tags?page_size=5" \
```
EvgeniyPatlan: Docker Hub's /v2/repositories/.../tags endpoint returns tags ordered by last_updated (most recently pushed first), not by version. So you're pulling the 5 most recently pushed tags and only then filtering/sorting locally. Two failure modes fall out of that:

After the strict ^3\.[0-9]+\.[0-9]+$ filter, those 5 tags can include latest, 3, 3.1, multiarch-suffixed tags, RCs, etc. — leaving fewer than 2 matches, so the pipeline errors out even though dozens of valid 3.x.y tags exist on the repo.
"Most recently pushed" ≠ "highest version." A rebuild or security re-push of an older tag (say 3.0.1) bumps its last_updated into your window of 5 while the actual newest 3.x.y falls outside it. sort -V then faithfully sorts a wrong/incomplete set and latestVersion is silently stale.

Fix is to fetch the full set (page_size up to 100, paginate via the response's next field) and do all filtering/sorting locally — never let the API's ordering stand in for "latest version."


Additionally 

A pipeline's exit status is the last command's (sort), so if curl fails on a rate-limit or transient 5xx, it emits nothing, jq reads empty input, and the pipeline "succeeds" with empty output. You then hit the < 2 branch with no hint that the real cause was the HTTP call. You want pipefail — but the Jenkins sh step runs under /bin/sh, which is dash on Debian/Ubuntu, and dash has no pipefail.

travagliad: Changed back to 100 page_size.

Changed the approach to run curl first so it will fail-fast.

@ademidoff tagging you so you know we're reverting page_size change, you have any inputs on that?

ademidoff (approved):

PR #4213, opened 2026-06-23
PG-2353: Auto-prune superseded AMIs + factory hardening

nogueiraanderson:
**Feature**
- After each prod promote the workflow prunes that (os, os_major, arch) combo down to the newest `KEEP_GENERATIONS=2` promoted bases plus the AMI pinned in `vars/moleculeEnvPPG.groovy` on master and everything newer than it, snapshots included, and `just prune-superseded` shares the same script for ad-hoc runs

**Why**
- `deprecate_at` only marks an AMI, it never deletes, so the inventory grows every bake (the live estate reached 105 promoted OL8/OL9 AMIs)
- The pins are static since [PR 4348](https://github.com/Percona-Lab/jenkins-pipelines/pull/4348) and lag the newest bake until the weekly refresh PR merges, so a keep-N prune alone would deregister the image pg.cd launches or the one an open refresh PR already names
- The prune is OS-scoped and guard-railed: exact `tag:os` whitelist, a bounded wait for the just-promoted AMI to become the newest visible match, keep-N floor, pin floor read from master at prune time with each of the 12 pins parsed exactly once, refusal when a candidate is pinned by any combo or named by an open refresh PR, prod bakes only, warn-and-continue loop with run annotations and step-summary reporting

**Tickets**
- [PG-2353](https://perconadev.atlassian.net/browse/PG-2353) (this), [PKG-1447](https://perconadev.atlassian.net/browse/PKG-1447) (static pins), companion [PR 4445](https://github.com/Percona-Lab/jenkins-pipelines/pull/4445) (dated refresh PRs and the daily pin check)

### DIFF
--- a/.github/workflows/ppg-ami-factory.yml
+++ b/.github/workflows/ppg-ami-factory.yml
@@ -47,6 +47,7 @@ concurrency:
 env:
   AWS_REGION: eu-central-1
   FACTORY_ENV: ${{ github.event.inputs.env || 'prod' }}
+  KEEP_GENERATIONS: '2' # per (os, os_major, arch) prod/test bases the post-promote prune keeps
 
 jobs:
   # No-AWS gate: a malformed template or bad combo fails HERE, before bake spins
@@ -100,10 +101,18 @@ jobs:
           aws-region: ${{ env.AWS_REGION }}
 
       - name: Install Session Manager plugin
+        env:
+          # Pinned to an immutable versioned URL + verified SHA256 (supply-chain): never
+          # `latest`. The plugin is installed in a job holding AWS OIDC creds, so the bytes
+          # are integrity-checked. Bump deliberately and re-record the digest of the new
+          # version's .deb (curl the versioned URL, sha256sum it).
+          SMP_VERSION: 1.2.835.0
+          SMP_SHA256: 7c6dcad12518571cc7959a713e6a8ae1bdf6ed66fd9bee37dc189e39ca58ae03
         run: |
           set -euo pipefail
           if ! command -v session-manager-plugin >/dev/null; then
-            curl -fsSL "https://s3.amazonaws.com/session-manager-downloads/plugin/latest/ubuntu_64bit/session-manager-plugin.deb" -o /tmp/smp.deb
+            curl -fsSL "https://s3.amazonaws.com/session-manager-downloads/plugin/${SMP_VERSION}/ubuntu_64bit/session-manager-plugin.deb" -o /tmp/smp.deb
+            echo "${SMP_SHA256}  /tmp/smp.deb" | sha256sum -c -
             sudo dpkg -i /tmp/smp.deb
           fi
           session-manager-plugin --version
@@ -158,8 +167,11 @@ jobs:
         run: |
           set -euo pipefail
           echo "smoke candidate: $AMI"
+          # `packer init` is CI infra (plugin download); a transient init failure must NOT
+          # deregister a good candidate. Only a real boot/install (build) failure does.
           # Deregister ONLY on a real boot/install failure (never on a later promote-tag failure).
-          ( cd smoke && packer init . && packer build -color=false \
+          ( cd smoke && packer init . ) || { echo "smoke packer init failed (CI infra, not a candidate defect); keeping $AMI"; echo "- smoke: init failed (kept $AMI)" >> "$GITHUB_STEP_SUMMARY"; exit 1; }
+          ( cd smoke && packer build -color=false \
               -var "candidate_ami=$AMI" -var "os=${OS}" -var "os_major=${OS_MAJOR}" \
               -var "arch=${ARCH}" -var "region=${AWS_REGION}" . ) \
           || { echo "smoke (boot+install) failed; deregistering $AMI + its snapshots"; aws ec2 deregister-image --delete-associated-snapshots --region "$AWS_REGION" --image-id "$AMI"; echo "- smoke: FAIL (deregistered $AMI)" >> "$GITHUB_STEP_SUMMARY"; exit 1; }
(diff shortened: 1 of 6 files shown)


ppg/packer/scripts/prune-superseded.sh
```diff
@@ -0,0 +1,81 @@
...
+  fi
+
+  if [[ "${APPLY}" != 1 ]]; then
+    echo "    would deregister ${ami} (superseded)"
+    continue
+  fi
+
+  echo "    deregister ${ami} (superseded)"
+
+  snaps=$(aws ec2 describe-images "${PROFILE_ARGS[@]}" --region "${REGION}" --image-ids "${ami}" \
```
EvgeniyPatlan: we already pass --delete-associated-snapshots in the smoke step. Why hand-roll describe->loop->delete-snapshot here? This orphans a snapshot if the delete fails after the image is already unregistered.

ppg/packer/scripts/prune-superseded.sh
```diff
@@ -0,0 +1,81 @@
...
+    --query 'Images[0].BlockDeviceMappings[].Ebs.SnapshotId' --output text) \
+    || { echo "    WARN describe ${ami} failed, skip" >&2; continue; }
+
+  aws ec2 deregister-image "${PROFILE_ARGS[@]}" --region "${REGION}" --image-id "${ami}"
```
EvgeniyPatlan: describe above is guarded, this isn't. One throttle on unregister and set -e bails on the rest of the combos. Wrap it || { echo WARN...; continue; } like you did the describe.

ppg/packer/justfile
```diff
@@ -259,25 +259,18 @@ prune-all apply="0":
...
-# AMI (the one the consumer selects) and remove older duplicates left by refreshes
-# (deprecate_at only marks, never deletes -> inventory grows). Lists by default; pass `1`
-# to deregister. NEVER deletes the newest-per-combo, and skips any AMI missing maj/arch.
+# deregister SUPERSEDED prod bases: per (os_major, arch) KEEP the newest `keep` role={{role_prod}}
+# AMIs (the consumer + refresh select the newest) and remove older duplicates left by refreshes
+# (deprecate_at only marks, never deletes -> inventory grows). Shares scripts/prune-superseded.sh
+# with the GHA workflow, which auto-prunes keep=2 after each promote. Lists by default; pass `1`.
 [doc("deregister superseded prod bases, keeping newest per combo; lists unless apply=1")]
-prune-superseded apply="0":
+prune-superseded apply="0" keep="1":
```
EvgeniyPatlan: workflow keeps 2, this defaults to 1. First person to run just prune-superseded 1 for cleanup wipes the rollback gen on every combo. Default keep="2" to match the automated path.

.github/workflows/ppg-oracle-ami-factory.yml
```diff
@@ -179,6 +190,19 @@ jobs:
...
+        continue-on-error: true # housekeeping must never fail an already-promoted bake
+        env:
+          OS_MAJOR: ${{ matrix.os_major }}
+          ARCH: ${{ matrix.arch }}
+        run: |
+          set -euo pipefail
+          # Deregister this combo's older promoted bases, keeping the newest KEEP_GENERATIONS.
+          # deprecate_at only marks, it never deletes, so without this the inventory grows.
+          role=ppg-package-test; [ "$FACTORY_ENV" = "test" ] && role=ppg-test-package-test
+          bash scripts/prune-superseded.sh "$role" "$OS_MAJOR" "$ARCH" "${KEEP_GENERATIONS:-2}" 1 "$AWS_REGION"
```
EvgeniyPatlan: continue-on-error + a silent exit 0 on empty + nothing written to the step summary = if this filter ever drifts, prune no-ops forever and we find out when the bill does. tee the output into $GITHUB_STEP_SUMMARY and emit a ::warning:: if the combo is still over keep after the run.

EvgeniyPatlan (changes requested):

nogueiraanderson:
Scope note: this prune is AWS-only. It speaks the EC2 API (`describe-images`, `deregister-image --delete-associated-snapshots`) and covers the AMIs the AWS factory promotes. The Hetzner factory in #4323 is not covered here and does not need to be: it ships its own retention prune (newest 4 promoted snapshots per combo after each bake) plus the hourly leak janitor, so both substrates cap their image inventory independently.

ppg/packer/scripts/prune-superseded.sh
```diff
@@ -0,0 +1,206 @@
...
+# (the signal that the filter drifted or failures piled up).
+residual=$(( ${#ami_ids[@]} - pruned_count ))
+
+if [[ "${APPLY}" == 1 ]]; then
+  summary "${COMBO}: pruned ${pruned_count}, failures ${prune_failures}, remaining ${residual} (keep ${KEEP_N})"
+
+  if (( residual > KEEP_N )); then
+    summary "::warning::${COMBO}: still ${residual} AMIs after pruning (over keep ${KEEP_N})"
+  fi
+fi
```
EvgeniyPatlan, line 215: the script has no non-zero exit path after the guards, so a combo where every DeregisterImage returns UnauthorizedOperation still exits 0. 
Add (( prune_failures > 0 || residual > KEEP_N )) && exit 1, otherwise the GHA step is green forever and the justfile's combo_failures counter never increments.

.github/workflows/ppg-ami-factory.yml
```diff
@@ -180,6 +192,31 @@ jobs:
           done
           echo "- promote: $pr" >> "$GITHUB_STEP_SUMMARY"
 
+      - name: Prune superseded (keep last ${{ env.KEEP_GENERATIONS }})
+        working-directory: ppg/packer
+        # Housekeeping must never fail an already-promoted bake, but it is not
+        # silent either: the script appends counts to $GITHUB_STEP_SUMMARY and
+        # emits ::warning:: annotations on any skipped deregister, failed
+        # snapshot cleanup, or combo left over the keep count.
+        continue-on-error: true
```
EvgeniyPatlan: continue-on-error: true also suppresses the notify job's failure() so even after the exit code is fixed this step still cannot page anyone

EvgeniyPatlan (changes requested):

EvgeniyPatlan (approved):

PR #4215, opened 2026-06-24
PG-2353: Shrink OL10 package-test AMI root to 20 GiB

nogueiraanderson:
**Bug**
- OL10 package-test AMIs shipped a 60 GiB root, which fails to launch on the PPG Molecule 30 GiB standard because EBS cannot restore a volume below its source snapshot (`InvalidBlockDeviceMapping`), blocking OL10 PPG jobs. OL8/OL9 are 20 GiB.

**Fix**
- Drop the OL10 `root_volume_size` special-case so OL10 refreshes at `var.volume_size` (20 GiB) like OL8/OL9.
- Add `reimage-ol10` + `reimage-ol10-verify`: a Packer `amazon-ebssurrogate` build (`reimage/reimage-ol10.pkr.hcl` + `scripts/reimage-surgery.sh`) copies the booted prebase root onto a smaller surrogate and registers the AMI from it, since EBS cannot restore below a snapshot and XFS cannot shrink in place. Fail-closed size gate, two-size boot + PPG smoke before promotion.
- Greenfield bootstrap promotes to a `prebase` role, and re-image shrinks it to the consumed role.

**Tickets**
- [PG-2353](https://perconadev.atlassian.net/browse/PG-2353) (this)

### DIFF
--- a/ppg/packer/justfile
+++ b/ppg/packer/justfile
@@ -25,6 +25,7 @@ os_name         := "oraclelinux"
 billing_tag     := "ppg-ami-factory"            # iit-billing-tag value (cost attribution + IAM fence)
 role_prod       := "ppg-package-test"           # promoted role the pg.cd consumer selects
 role_test       := "ppg-test-package-test"      # isolated promoted role for env=test bakes
+role_prebase    := "ppg-ol10-prebase"           # greenfield bootstrap output; reimage shrinks it before promotion
 subnet          := "subnet-068170595951ab3a9"   # eu-central-1 default-VPC public subnet
 bucket_prefix   := "ppg-ami-factory-import"     # import bucket = "{{bucket_prefix}}-<account-id>"
 import_role     := "ppg-ol10-vmimport"          # vmimport role for the OL10 import-snapshot
@@ -35,6 +36,7 @@ boottest_sg     := "ppg-ol10-boottest-sg"       # OL10 boot-test SSH security gr
 # --- build matrix dimensions ---
 os_majors       := "8 9 10"
 arches          := "x86_64 arm64"
+root_gib        := "20"                          # reimage target root GiB; MUST match var.volume_size in oracle-linux.pkr.hcl
 
 # common create-tags / role-tags set (billing only), reused everywhere
 common_tags     := "Key=iit-billing-tag,Value=" + billing_tag
@@ -94,7 +96,7 @@ _prune-by-filter apply +filters:
     [ -z "$ids" ] && { echo "  (nothing matches)"; exit 0; }
     for ami in $ids; do
       role=$(aws ec2 describe-images --profile {{profile}} --region {{region}} --image-ids "$ami" \
-        --query 'Images[0].Tags[?Key==`role`]|[0].Value' --output text)
+        --query "Images[0].Tags[?Key=='role']|[0].Value" --output text)
       # fail-closed: never delete the prod base, nor anything we cannot positively
       # classify (missing/None role) -- "don't delete what you can't identify".
       if [ "$role" = "{{role_prod}}" ] || [ -z "$role" ] || [ "$role" = None ]; then echo "  SKIP $ami (protected: role=${role:-<none>})"; continue; fi
@@ -111,10 +113,12 @@ _prune-by-filter apply +filters:
 init:
     packer init .
 
-# format the HCL (main + smoke templates)
+# format the HCL (main + smoke + reimage + bootstrap templates)
 fmt:
     packer fmt .
     packer fmt smoke
+    packer fmt reimage
+    packer fmt bootstrap
 
 # validate one combo  (maj=8|9|10  arch=x86_64|arm64)
 validate maj="9" arch="x86_64":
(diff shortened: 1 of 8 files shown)


ppg/packer/scripts/reimage-surgery.sh
```diff
@@ -0,0 +1,123 @@
...
+
+ARCH="${ARCH:?set ARCH}"
+T="${TGT:?set TGT (target disk, e.g. /dev/nvme1n1)}"
+
+command -v parted >/dev/null || dnf -y install parted
+command -v rsync  >/dev/null || dnf -y install rsync
+[ "$ARCH" = arm64 ] && { command -v mkfs.fat >/dev/null || dnf -y install dosfstools; }
+for t in mkfs.xfs mkswap partprobe xfs_freeze blkid findmnt; do
+  command -v "$t" >/dev/null || { echo "MISSING required tool: $t" >&2; exit 1; }
+done
```
EvgeniyPatlan, line 55: why not to fail as early as possible on dnf install stage? if exit code 1 why we need to wait?

ppg/packer/scripts/reimage-ol10.sh
```diff
@@ -0,0 +1,149 @@
...
+HERE="$(cd "$(dirname "$0")" && pwd)"
+
+case "$ARCH" in
+  arm64)  EC2ARCH=arm64;  INST=t4g.large; BOOTMODE=uefi ;;
+  x86_64) EC2ARCH=x86_64; INST=t3.large;  BOOTMODE=legacy-bios ;;
+  *) echo "ARCH must be arm64 or x86_64" >&2; exit 1 ;;
+esac
+if [ "$ENV" = test ]; then CAND_ROLE="ppg-reimage-test-candidate"; SRC_TAG="factory-test"; F_ENV="test"; NPFX="TEST-OL"
+else CAND_ROLE="ppg-reimage-candidate"; SRC_TAG="factory"; F_ENV="prod"; NPFX="OL"; fi
+TS="$(date -u +%Y%m%d-%H%M%S)"; NAME="${NPFX}10-${ARCH}-reimage-${TS}"
```
EvgeniyPatlan: NAME is second-resolution and it's also the recovery-tag value used in cleanup() (line 74). two same-arch runs in the same second share it, and then one run's cleanup finds the other's instance/volume/snapshot by tag and deletes them. verify already uses {{ami}}-$$ for exactly this, append $$ here too.

ppg/packer/justfile
```diff
@@ -392,6 +399,76 @@ bootstrap-ol10-verify ami arch="arm64":
...
+# boot-validate a re-imaged candidate on TWO root sizes ({{root_gib}} + 30) + smoke, then promote.
+# Fail-closed SIZE GATE: never promote a base whose root snapshot exceeds {{root_gib}}
+# (= var.volume_size), or the refresh could not launch it (EBS won't restore below the snapshot).
+reimage-ol10-verify ami arch="arm64" env="prod":
+    #!/usr/bin/env bash
+    set -euo pipefail
+    eval "$(aws configure export-credentials --profile {{profile}} --format env)"; unset AWS_PROFILE
+    gib=$(aws ec2 describe-images --region {{region}} --image-ids {{ami}} \
+      --query 'Images[0].BlockDeviceMappings[0].Ebs.VolumeSize' --output text)
+    [ "$gib" -le "{{root_gib}}" ] || { echo "REFUSE promote: {{ami}} root is ${gib} GiB > {{root_gib}} GiB target"; exit 1; }
```
EvgeniyPatlan, line 448: the candidate AMI + snapshot from reimage-ol10 never get cleaned up if verify fails. this gate exits before the trap is even set at 453, and the trap only terminates instances anyway. deregister the candidate on the failure paths here.

ppg/packer/scripts/reimage-ol10.sh
```diff
@@ -0,0 +1,149 @@
...
+[ -n "$SRC" ] || { echo "no $ROLE_PROMOTED or $ROLE_PREBASE OL10 $ARCH base; bootstrap one first"; exit 1; }
+CUR=$(snap_gib "$SRC")
+if [ "$FROM" = consumed ] && [ "$CUR" -le "$ROOT_GIB" ] && [ "$FORCE" != 1 ]; then
+  echo "OL10 $ARCH consumed base $SRC root is ${CUR} GiB (<= ${ROOT_GIB}); already shrunk. Nothing to do (FORCE=1 to re-run)."
+  exit 0
+fi
+echo "=== re-imaging OL10 $ARCH $FROM base $SRC (${CUR} GiB) -> ${ROOT_GIB} GiB (env=$ENV) ==="
+
+# 1. launch a builder FROM the base; SSH in via the boot-test key.
+sg=$(aws ec2 describe-security-groups --region "$REGION" --filters Name=group-name,Values="$BOOTTEST_SG" --query 'SecurityGroups[0].GroupId' --output text)
```
EvgeniyPatlan: nothing stops a second run starting while a builder is already up. ok if we treat this as serial-only but that's not written down or enforced anywhere. could add a quick describe-instances on the builder Name tag + arch and bail if one's pending/running.

ppg/packer/scripts/reimage-surgery.sh
```diff
@@ -0,0 +1,123 @@
...
+  parted -s "$T" mkpart swap linux-swap 1225MiB 5321MiB
+  parted -s "$T" mkpart root xfs        5321MiB 100%
+  ESP_P=${T}p1
+else
+  parted -s "$T" mkpart bios_grub 1MiB 3MiB ; parted -s "$T" set 1 bios_grub on
+  parted -s "$T" mkpart boot xfs        3MiB    1027MiB
+  parted -s "$T" mkpart swap linux-swap 1027MiB 5123MiB
+  parted -s "$T" mkpart root xfs        5123MiB 100%
+fi
+BOOT_P=${T}p2; SWAP_P=${T}p3; ROOT_P=${T}p4
```
EvgeniyPatlan: the p suffix only works because TGT is always nvme (t3/t4g are nitro). if anyone changes the builder instance family and the device comes up as /dev/sdf, ${T}p2 becomes /dev/sdfp2 and surgery fails in a confusing way. derive it: append p only when the device name ends in a digit.

ppg/packer/justfile
```diff
@@ -392,6 +399,76 @@ bootstrap-ol10-verify ami arch="arm64":
...
+    # cross-env guard: never promote a candidate whose factory_env tag does not match {{env}}, so a
+    # test candidate can never be promoted to the production role by an omitted/wrong env arg.
+    fenv=$(aws ec2 describe-images --region {{region}} --image-ids {{ami}} --query 'Images[0].Tags[?Key==`factory_env`]|[0].Value' --output text)
```
EvgeniyPatlan: the backtick form Key==factory_env`` leans on jmespath being lenient about it. single quotes (Key=='factory_env') is the portable form and won't break if the cli's jmespath gets bumped.

ppg/packer/justfile
```diff
@@ -392,6 +399,76 @@ bootstrap-ol10-verify ami arch="arm64":
...
+        --query 'Instances[0].InstanceId' --output text)
+      iids="$iids $iid"
+      echo "launched $iid (root=${sz}GiB); waiting 2/2..."; aws ec2 wait instance-status-ok --region {{region}} --instance-ids "$iid"
+      ip=$(aws ec2 describe-instances --region {{region}} --instance-ids "$iid" --query 'Reservations[0].Instances[0].PublicIpAddress' --output text)
+      min=$(( sz - ovh_gib ))   # after growpart the root fs >= launch size minus the derived overhead
+      ssh -i ol10-import/{{boottest_key}}.pem $so ec2-user@"$ip" "rg=\$(findmnt -bno SIZE /); rg=\$((rg/1073741824)); echo root=\${rg}G; [ \$rg -ge $min ] || { echo \"FAIL growpart: root \${rg}G < ${min}G on a ${sz}G launch\"; exit 1; }; sudo OS_MAJOR=10 UNAME_ARCH=$uarch bash -s" < scripts/validate.sh
+    done
+    pr={{role_prod}}; [ "{{env}}" = "test" ] && pr={{role_test}}
+    echo "smoke (boot+install) {{ami}} -> promote role=$pr"
+    just smoke {{ami}} 10 {{arch}} $pr
```
EvgeniyPatlan, line 488: since _promote only adds the role tag and never strips it from the old base, after this runs both the 60G and 20G AMIs carry role=ppg-package-test. does the consumer resolve by newest CreationDate, or first-match? if it's first-match this looks like the fix didn't take. worth confirming before we promote, since that's the most likely false-failure.

EvgeniyPatlan (changes requested):
The code is overloaded with AI comments that were added during the code generation. very hard to read.
Also please improve code format wherever is possible as readability is low

EvgeniyPatlan (approved):

PR #4248, opened 2026-07-10
PMM: build pmm-server for arm64

EvgeniyPatlan:
This adds a native arm64 build of pmm-server v3, following the same layout we already use for the client:

- pmm3-server-autobuild becomes a parent job that runs two new jobs (pmm3-server-autobuild-amd and pmm3-server-autobuild-arm) in parallel and merges their images into a multi-arch manifest with docker buildx imagetools. It still exports TIMESTAMP_TAG, so pmm3-release-candidate doesn't need any changes.
- watchtower is built for both arches too - the binary cross-compiles, the image is assembled per arch and merged the same way. It runs next to the server on every staging instance, so an arm server host can't live without it.
- pmm3-api-tests got an AGENT_ARCH parameter and the server parent now triggers it for both arches with the image it just built.
- pmm3-image-scanning scans both platforms of the manifest and archives per-arch reports.
- pmm3-aws-staging-start got a SERVER_ARCH parameter (arm64 = t4g.xlarge). runSpotInstance takes the arch as a second argument, default is x86_64 so nothing changes for existing callers. The arm64 worker AMI already exists (same pmm-worker-3 tag, the packer template builds both since a while).
- package testing jobs and their matrices pass SERVER_ARCH through, so the arm matrix now runs client and server both on arm.
- setupPMM3Client: latest-tarball resolves to the arm client tarball on aarch64 hosts, before it always downloaded the amd64 one.

All of this was tested from a test branch with separate -test job copies and -test docker tags before opening this PR: the multi-arch manifest builds, api tests are green on both arches, an arm64 staging instance comes up with server+watchtower+client working, package tests ran against it.

Depends on percona/pmm#5630 - without the arch-namespaced S3 build cache the arm build downloads cached x86_64 rpms and skips building. Merge that one first.

Two new jobs need to be registered when this merges: pmm3-server-autobuild-amd and pmm3-server-autobuild-arm (pipeline from SCM, master, script paths match the file names).

Not in this PR, follow-ups:
- pmm3-client-repo-push uploads both arch tarballs to TESTING/pmm/ under the same file name, so amd64 and arm64 overwrite each other on every RC. Needs an arch-aware destination.
- rc-testing lanes and nightly UI runs against an arm server.
- arm64 AMI wiring in pmm3-ami (the packer side is already in percona/pmm#5630).

### DIFF
--- a/pmm/v3/pmm3-watchtower-autobuild.groovy
+++ b/pmm/v3/pmm3-watchtower-autobuild.groovy
@@ -72,14 +72,17 @@ pipeline {
                 }
             }
         }
-        stage('Build watchtower binary') {
+        stage('Build watchtower binaries') {
             steps {
                 withCredentials([[$class: 'AmazonWebServicesCredentialsBinding', accessKeyVariable: 'AWS_ACCESS_KEY_ID', credentialsId: 'pmm-staging-slave', secretKeyVariable: 'AWS_SECRET_ACCESS_KEY']]) {
                     sh '''
-                        docker run --rm \
-                            -v ${WORKSPACE}/${PATH_TO_WATCHTOWER}:/watchtower \
-                            public.ecr.aws/e7j3v3n0/rpmbuild:3 \
-                            make -C /watchtower build
+                        for arch in amd64 arm64; do
+                            docker run --rm \
+                                -v ${WORKSPACE}/${PATH_TO_WATCHTOWER}:/watchtower \
+                                -e GOARCH=${arch} \
+                                public.ecr.aws/e7j3v3n0/rpmbuild:3 \
+                                bash -c "make -C /watchtower build && mv /watchtower/watchtower /watchtower/watchtower-${arch}"
+                        done
                     '''
                 }
             }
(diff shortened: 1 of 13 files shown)


pmm/v3/pmm3-watchtower-autobuild.groovy
```diff
@@ -95,17 +98,32 @@ pipeline {
...
                     cd ${PATH_TO_WATCHTOWER}
-                    docker build -t ${TIMESTAMP_TAG} -f dockerfiles/Dockerfile .
-                    docker push ${TIMESTAMP_TAG}
 
+                    docker buildx create --name multiarch-wt --driver docker-container --use || docker buildx use multiarch-wt
+
+                    for arch in amd64 arm64; do
+                        rm -rf ctx-${arch}
+                        mkdir -p ctx-${arch}/dockerfiles
+                        cp watchtower-${arch} ctx-${arch}/watchtower
```
ademidoff, line 110: Let's remove everything related to Watchtower as it is being discontinued.

EvgeniyPatlan: @ademidoff it is just test job so it really now copies the old behavior. Sure we can remove this job. But I suggest keeping it for now so it would be possible to test with and without watchtower,

PR #4262, opened 2026-07-16
PG-2353: Auto-prune superseded OL AMIs safely after each promote

nogueiraanderson:
**Feature**
- After each promote the factory keeps the newest `KEEP_GENERATIONS` (2) AMIs per (os_major, arch) and deregisters older ones together with their snapshots

**Why**
- `deprecate_at` only marks an AMI deprecated, it never deletes, so the inventory grows ~6 AMIs per weekly bake
- Deletion automation next to production consumers must fail safe: native newest-first selection with server-side architecture filters, warn-and-continue on throttling, `role=*superseded*` demotions and mistagged AMIs never match, and a promoted-AMI guard refuses to prune during the EC2 eventual-consistency window
- The justfile helpers go native too: `--delete-associated-snapshots` replaces the hand-rolled snapshot loops (fixing an unreachable already-gone path) and the per-AMI re-describes collapse into single calls

**Tickets**
- [PG-2353](https://perconadev.atlassian.net/browse/PG-2353) (this), supersedes [#4213](https://github.com/Percona-Lab/jenkins-pipelines/pull/4213) with its review addressed

### DIFF
--- a/.github/workflows/ppg-oracle-ami-factory.yml
+++ b/.github/workflows/ppg-oracle-ami-factory.yml
@@ -41,6 +41,7 @@ concurrency:
 env:
   AWS_REGION: eu-central-1
   FACTORY_ENV: ${{ github.event.inputs.env || 'prod' }}
+  KEEP_GENERATIONS: '2' # per (os_major, arch) prod/test bases the post-promote prune keeps
 
 jobs:
   # No-AWS gate: a malformed template or bad combo fails HERE, before bake spins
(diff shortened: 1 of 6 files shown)


nogueiraanderson:
#### Review mapping (from #4213)

- Hand-rolled snapshot loop replaced with `deregister-image --delete-associated-snapshots`; non-success `DeleteSnapshotResults` codes surface as warnings
- Unguarded deregister now warns and continues, one throttled call no longer aborts the remaining AMIs or combos
- `keep` defaults aligned at 2 in the justfile and the workflow, `KEEP_N=0` stays rejected
- The prune step reports counts to `$GITHUB_STEP_SUMMARY` and emits `::warning::` on every skipped deregister, failed snapshot cleanup, or combo left over the keep count

Additional hardening beyond the original review: pagination-safe listing (plain JSON fetched once, sorted locally), `prune-stale` skips `role=*superseded*` demotions, native-architecture mismatches are excluded and warned, and a `PROMOTED_AMI` guard refuses to prune while the just-promoted AMI is not yet the newest visible match.

#### Validation

- Statics clean: shellcheck (Google-aligned rc), actionlint, justfile parse, workflow YAML parse
- Eight mocked scenarios against a fake `aws` (shuffled-date sorting, apply set, empty estate, mid-loop throttle, promoted-AMI mismatch, cross-architecture imposter, partial snapshot failure, `keep=0`), all assertions on output plus the recorded mutating-call log
- Live dry-run across all six (os_major, arch) combos in eu-central-1: keeps the two newest weekly generations per combo, proposes 26 stale AMIs for cleanup, and the two demoted `oversized-superseded` OL10 images are never candidates

nogueiraanderson:
Folded back into the reopened [#4213](https://github.com/Percona-Lab/jenkins-pipelines/pull/4213), which now carries this exact head with the original review threads resolved.

PR #4276, opened 2026-07-17
 K8SPSMDB-1601 fixes for Rancher pipeline

valmiranogueira:
- Fix min and max RKE2 version from release_versions file
- Print test variables before start test in order to check processed variables from release versions
- Improve job description by add arch and process DB_VERSION variable


cloud/jenkins/psmdbo_rancher.groovy
```diff
@@ -24,12 +24,12 @@ pipeline {
...
         text(name: 'TEST_LIST', defaultValue: '', description: 'List of tests to run separated by new line')
         choice(name: 'IGNORE_PREVIOUS_RUN', choices: ['NO', 'YES'], description: 'Ignore passed tests in previous run')
-        choice(name: 'PILLAR_VERSION', choices: ['none', '80', '70', '60'], description: 'Implies release run.')
-        string(name: 'GIT_BRANCH', defaultValue: 'main', description: 'Tag/Branch')
-        choice(name: 'PLATFORM_CHANNEL', choices: ['stable', 'latest', 'testing'], description: '')
-        string(name: 'PLATFORM_VERSION', defaultValue: 'latest', description: 'Kubernetes version')
-        string(name: 'PLATFORM_ARCH', defaultValue: 'amd64', description: 'Platform architecture')
-        string(name: 'RANCHER_VERSION', defaultValue: 'latest', description: 'Rancher chart version')
+        choice(name: 'PILLAR_VERSION', choices: ['none', '80', '70', '60'], description: 'Set to 60/70/80 for a release run. Release runs force PLATFORM_CHANNEL=stable and load certified image tags from source/e2e-tests/release_versions.')
```
jvpasinatto: we use certified in another context so maybe not best word here

cloud/common/vars/tests.groovy
```diff
@@ -49,72 +114,107 @@ String getReleaseParamName(String imageName, String pillarVersion, String operat
...
+}
 
-            default:
-                error("Unsupported platform_provider: ${testVariables.platform_provider}")
-        }
+String getPlatformReleasePrefix(Map testVariables) {
+    def platformPrefix = [
+        "gcloud"      : "GKE",
+        "azure"       : "AKS",
+        "redhat"      : "OPENSHIFT",
```
jvpasinatto: why use redhat here?  lets just use openshift if possible

jvpasinatto (changes requested):

cloud/common/vars/tests.groovy
valmiranogueira: I'm using provider names like google cloud (gcloud), azure, redhat... It's weird to say that doks is digital ocean, and gke is gcloud, but openshift is openshift if that makes sense.
```
def platformPrefix = [
        "gcloud"      : "GKE",
        "azure"       : "AKS",
        "redhat"      : "OPENSHIFT",
        "digitalocean": "DOKS",
        "rancher"     : "RKE2",
    ][testVariables.platform_provider?.toLowerCase()]
```
    
'platform_provider', only Rancher that is not a cloud provider, but k8s instances provider. It's related Kubernetes is RKE2.

In case of minikube, it can be minikube cause there is no platform provider.

At the same time platform_provider can change to platform type, but then we need to remember that rke2 is rancher.

cloud/jenkins/psmdbo_rancher.groovy
valmiranogueira: Fixed

cloud/common/vars/tests.groovy
valmiranogueira: fixed

jvpasinatto (approved):

PR #4288, opened 2026-07-22
PKG-1342: Extend the PPG AMI factory to bake Rocky Linux 8/9/10

nogueiraanderson:
**Feature**
- The PPG package-test AMI factory now bakes Rocky Linux 8/9/10 (x86_64 + arm64) alongside Oracle Linux: same refresh pipeline, fail-closed validate, fresh-boot PPG smoke, and tag-flip promotion.
- The workflow is renamed `ppg-ami-factory.yml` and runs its `check` job (template validation only, no AWS credentials, so nothing can launch or spend) on pull requests.

**Why**
- The pg.cd molecule jobs pin static Rocky AMIs. The Rocky 8 pin (`8.10-20240528`) is AWS-deprecated and two years behind on errata, and Rocky's June 2026 refresh reached eu-central-1 only as a Marketplace image whose product code would propagate into derived AMIs.
- Each Rocky base image is seeded once from the official community AMI (`just seed-rocky`, `include_deprecated` covers Rocky 8), then refreshed weekly like OL, keeping Rocky 8 current even when upstream skips our region.
- Rocky is in the weekly scheduled rebake from this PR on. The initial Rocky images are already baked, smoke-passed and promoted, so every scheduled run has a source AMI to build from (each rebake selects the newest promoted image as its source).
- The Rocky jobs keep launching the old hand-pinned AMIs for now. The follow-up PR points them at the factory images instead (`moleculeEnvPPG.groovy` resolves the newest promoted AMI by tag in place of six hardcoded ids).

**Tickets**
- [PKG-1343](https://perconadev.atlassian.net/browse/PKG-1343) (Rocky 8), [PKG-1344](https://perconadev.atlassian.net/browse/PKG-1344) (Rocky 9), [PKG-1345](https://perconadev.atlassian.net/browse/PKG-1345) (Rocky 10), under epic [PKG-1342](https://perconadev.atlassian.net/browse/PKG-1342)

[PKG-1343]: https://perconadev.atlassian.net/browse/PKG-1343?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/ppg-ami-factory.yml
+++ b/.github/workflows/ppg-ami-factory.yml
@@ -1,22 +1,24 @@
-# Oracle Linux package-test AMI factory.
-# Bakes OL8/OL9 (and later OL10) x86_64+arm64 AMIs with Packer over AWS Session
+# EL package-test AMI factory: Oracle Linux + Rocky Linux.
+# Bakes OL and Rocky 8/9/10 x86_64+arm64 AMIs with Packer over AWS Session
 # Manager (no inbound SSH), authenticated via GitHub OIDC (no static keys), and
 # promotes each via a fresh-boot smoke test. Consumer: pg.cd molecule jobs select
 # the newest role=ppg-package-test AMI by tag (vars/moleculeEnvPPG.groovy).
+# Rocky lineage roots are seeded once per combo via dispatch (just ci-seed-rocky).
 #
-# Triggers: recipe change (paths), weekly security rebake (cron), manual.
+# Triggers: PR validate-only (pull_request -> check job, no AWS), recipe change
+# (paths), weekly security rebake (cron), manual.
 # Actions are pinned to commit SHAs; the trailing "# vX.Y.Z" records the tag.
 
-name: ppg-oracle-ami-factory
-run-name: ppg OL AMI factory (${{ inputs.env || 'prod' }}, ${{ github.event_name }})
+name: ppg-ami-factory
+run-name: ppg AMI factory (${{ inputs.env || 'prod' }}, ${{ github.event_name }})
 
 on:
   workflow_dispatch:
     inputs:
       combos:
-        description: 'JSON matrix include list, e.g. [{"os_major":"9","arch":"x86_64"}]'
+        description: 'JSON matrix include list, e.g. [{"os":"rocky","os_major":"9","arch":"x86_64","seed":true}]; os defaults to oraclelinux, seed to false'
         required: false
-        default: '[{"os_major":"9","arch":"x86_64"}]'
+        default: '[{"os":"oraclelinux","os_major":"9","arch":"x86_64"}]'
       env:
         description: 'prod (real factory tags) or test (isolated tags, never consumed by pg.cd)'
         required: false
@@ -27,9 +29,13 @@ on:
     branches: [master]
     paths:
       - 'ppg/packer/**'
-      - '.github/workflows/ppg-oracle-ami-factory.yml'
+      - '.github/workflows/ppg-ami-factory.yml'
+  pull_request:
+    paths:
+      - 'ppg/packer/**'
+      - '.github/workflows/ppg-ami-factory.yml'
   schedule:
-    - cron: '0 6 * * 1' # weekly Mon 06:00 UTC - absorb Oracle security errata
+    - cron: '0 6 * * 1' # weekly Mon 06:00 UTC - absorb OL + Rocky security errata
 
 permissions:
   contents: read # workflow-level default: read-only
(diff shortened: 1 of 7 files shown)


.github/workflows/ppg-ami-factory.yml
```diff
@@ -27,9 +29,13 @@ on:
     branches: [master]
     paths:
       - 'ppg/packer/**'
-      - '.github/workflows/ppg-oracle-ami-factory.yml'
+      - '.github/workflows/ppg-ami-factory.yml'
+  pull_request:
+    paths:
+      - 'ppg/packer/**'
+      - '.github/workflows/ppg-ami-factory.yml'
```
EvgeniyPatlan, line 36: The PR trigger fires on ppg/packer/**, which includes reimage/ and bootstrap/, but the check job below never validates those templates.

ppg/packer/scripts/validate.sh
```diff
@@ -1,45 +1,60 @@
...
 
-# 1. Genuine Oracle Linux of the expected major (fidelity gate).
-[ -f /etc/oracle-release ] || fail "no /etc/oracle-release (not Oracle Linux)"
-grep -Eq "release[[:space:]]+${OS_MAJOR}\." /etc/oracle-release \
-  || fail "oracle-release major != ${OS_MAJOR}: $(cat /etc/oracle-release)"
+# 1. Genuine OS of the expected major (fidelity gate).
+case "${OS}" in
+  oraclelinux) release_file=/etc/oracle-release ;;
+  rocky)       release_file=/etc/rocky-release ;;
+  *)           fail "unknown OS '${OS}' (want oraclelinux|rocky)" ;;
```
EvgeniyPatlan: it is good for now but if we are going to extend and add redhat - it will fail as on all oracle/rocky/alma linuxes /etc/redhat-release is present. Better not to check filename and decide what system is it but to check context of the file.

ppg/packer/scripts/validate.sh
```diff
@@ -1,45 +1,60 @@
...
+  crb="ol${OS_MAJOR}_codeready_builder"
+  dnf repolist --all 2>/dev/null | grep -qiE "codeready" \
+    || fail "CRB repo (${crb}) not defined on image"
+else
+  crb=crb
+  [[ "${OS_MAJOR}" = 8 ]] && crb=powertools
+  dnf repolist --all 2>/dev/null | grep -qiE "^${crb}([[:space:]]|$)" \
+    || fail "CRB repo (${crb}) not defined on image"
+fi
+echo "  CRB ok: repo defined (${crb})"
```
EvgeniyPatlan: there is a code duplication and we will just add more if in case we need to extend it. maybe we can use something like this:
```
case "${OS}/${OS_MAJOR}" in
  oraclelinux/*) crb="ol${OS_MAJOR}_codeready_builder" ;;
  rocky/8)       crb="powertools" ;;
  rocky/*)       crb="crb" ;;
esac
dnf repolist --all 2>/dev/null | grep -qiE "^${crb}([[:space:]]|$)" \
  || fail "CRB repo (${crb}) not defined on image"
echo "  CRB ok: repo defined (${crb})"
```

ppg/packer/smoke/smoke.pkr.hcl
```diff
@@ -43,17 +51,22 @@ variable "builder_security_group_name" {
 locals {
   instance_type = var.arch == "arm64" ? "t4g.large" : "t3.large"
   uname_arch    = var.arch == "arm64" ? "aarch64" : "x86_64"
+  release_file  = var.os == "rocky" ? "/etc/rocky-release" : "/etc/oracle-release"
```
EvgeniyPatlan: again release file would file in case we extend and add redhat

EvgeniyPatlan (changes requested):

.github/workflows/ppg-ami-factory.yml
nogueiraanderson: Good catch. The check job now fmt-checks and validates the reimage and bootstrap templates too, matching the local `just check`.

ppg/packer/scripts/validate.sh
nogueiraanderson: Agreed. The gate now keys on `/etc/os-release` `ID` + `VERSION_ID` major (`ol` / `rocky`), and rhel or almalinux slot in as new case arms. Verified against OL 8/9/10 and Rocky 8/9/10 userlands plus a negative case.

ppg/packer/scripts/validate.sh
nogueiraanderson: Applied as suggested, with a fail-closed default arm. The anchored grep also matches the OL ids exactly (`ol<major>_codeready_builder`), verified on OL 8/9/10.

ppg/packer/smoke/smoke.pkr.hcl
nogueiraanderson: Same fix as validate.sh: the smoke identity lines now read `/etc/os-release` `ID` + `VERSION_ID` major.

EvgeniyPatlan (approved):

PR #4319, opened 2026-08-04
CLOUD-934 Use common functions for PXC Operator pipelines

jvpasinatto:
- Use common functions to simplify and reduce repeated code across PXC release pipelines


valmiranogueira (approved):
Nice! Mind  I also have an example for minikube that I was working on since newer versions of minikube tests were failing for PG: https://github.com/Percona-Lab/jenkins-pipelines/tree/use-libraries-for-minikube.

Perhaps we can merge this one now and I can add the shared minikube cluster creation later too :)

PR #4328, opened 2026-08-07
CLOUD-934 Use common functions for PS Operator pipelines

jvpasinatto:
Clean PS pipelines and use common functions


cloud/jenkins/pso_rancher.groovy
```diff
@@ -100,8 +92,7 @@ pipeline {
                 script {
                     libraries.tools.dockerBuildAndPush(
                         operatorImage: operatorImage,
-                        branch: GIT_BRANCH,
-                        operator: 'ps-operator'
+                        branch       : GIT_BRANCH
```
eleo007, line 95: Do we remove  operator: 'ps-operator' bc we need it only for "pg-operator" ?https://github.com/Percona-Lab/jenkins-pipelines/blob/eef0dee87964692876a25505d7b190403318dd70/cloud/common/vars/tools.groovy#L121

jvpasinatto: yes. and the aim is to make every operator use make build, but for now i opted by the way PR jenkins job uses it

eleo007 (approved):

cloud/common/vars/dependencies.groovy
```diff
@@ -204,12 +212,28 @@ void installProviderDependencies(Map libraries, String operator, String provider
         installAzureCLI()
         libraries.azure.auth()
     }
+
+    if (provider == 'eks') {
```
valmiranogueira, line 216: Nit: switch case for this makes it less repetitive and only one condition is executed

cloud/common/vars/tests.groovy
```diff
@@ -780,6 +820,30 @@ void clusterRunner(String clusterSuffix, Map testVariables) {
     }
 }
 
+void clusterRunnerWithProviderCredentials(String clusterSuffix, Map testVariables) {
+    def provider = testVariables.platform_provider.toLowerCase()
+
+    if (provider == "eks") {
```
valmiranogueira, line 826: Same nit here

PR #4329, opened 2026-08-07
EC2 instance checks add stop and termination for PGSQL

panchal-yash:


kaushikpuneet07 (approved):
LGTM

misc/ec2-instance-checks.groovy
```diff
@@ -22,125 +22,246 @@ pipeline {
...
-                            env.ovqa = sh(script: "cat ${WORKSPACE}/overview-qa.txt",returnStdout: true).trim()
-                            echo "Print the overview qa after removing the 0 servers list in QA \n ${env.ovqa}"
-                            env.ovqacount=sh(script: """ awk '{sum += \$4} END {print sum}' ${WORKSPACE}/overview-qa.txt """,returnStdout: true).trim()
-             }
 
+
+                        withCredentials([[$class: 'AmazonWebServicesCredentialsBinding', accessKeyVariable: 'AWS_ACCESS_KEY_ID', credentialsId: 'c42456e5-c28d-4962-b32c-b75d161bff27', secretKeyVariable: 'AWS_SECRET_ACCESS_KEY']]) {
+
+                            // Runs every 6 hrs (every build) — PGSQL checks
+                            sh "bash -x ./check-ec2-instances-pgsql.sh"
```
nogueiraanderson, line 54: check-ec2-instances-pgsql.sh terminates any running instance whose iit-billing-tag contains jenkins-pg-worker after 6h uptime. That's the exact tag pg.cd's production build workers carry (cloud.groovy:175 in cd-platform), not just the molecule test VMs. A production build running past 6h gets terminate-instances'd mid-build.

misc/ec2-instance-checks.groovy
```diff
@@ -22,125 +22,246 @@ pipeline {
...
-                            echo "Instance ${id} in region ${regionForId}"
-                            echo "Preparing to terminate instance: ${id} in region: ${regionForId}"
-                            sh "aws ec2 terminate-instances --instance-ids ${id} --region ${regionForId}"
-                            echo "Instance ${id} in region ${regionForId} terminated."
+                stopLines.each { line ->
+                    def regionMatch = line =~ /-+Region\s+([a-z0-9-]+)\s+has/
+                    if (regionMatch) {
+                        pgsqlStopRegion = regionMatch[0][1]
+                    }
+                    def instanceMatch = line =~ /(i-[a-zA-Z0-9]+)/
```
nogueiraanderson, line 150: The stop selection has no tag check at all, just KeyName in mohit.joshi-keyeu1, shahid-key, naeem. Any running instance under those three keys gets stopped every 6h regardless of what it's for.

misc/ec2-instance-checks.groovy
```diff
@@ -22,125 +22,246 @@ pipeline {
...
+                    if (instanceMatch) {
+                        def instanceId = instanceMatch[0][1]
+                        pgsqlStopPairs << [id: instanceId, region: pgsqlStopRegion]
+                    }
+                }
+
+                withCredentials([[$class: 'AmazonWebServicesCredentialsBinding', accessKeyVariable: 'AWS_ACCESS_KEY_ID', credentialsId: 'c42456e5-c28d-4962-b32c-b75d161bff27', secretKeyVariable: 'AWS_SECRET_ACCESS_KEY']]) {
+
+                    // Terminate PGSQL instances (>6 hrs)
+                    pgsqlTerminatePairs.each { pair ->
```
nogueiraanderson, line 160: terminate-instances runs in a plain .each, so one failed call (already-terminated, throttled, wrong region) throws and skips every remaining pair, the whole stop loop below, and every Slack notification. One bad instance shouldn't be able to swallow the rest silently.

misc/ec2-instance-checks.groovy
panchal-yash: It is a requirement by them in case they forget to stop the instance after testing manually.

misc/ec2-instance-checks.groovy
panchal-yash: @Naeem-Akhter

misc/ec2-instance-checks.groovy
```diff
@@ -22,125 +22,261 @@ pipeline {
...
+                withCredentials([[$class: 'AmazonWebServicesCredentialsBinding', accessKeyVariable: 'AWS_ACCESS_KEY_ID', credentialsId: 'c42456e5-c28d-4962-b32c-b75d161bff27', secretKeyVariable: 'AWS_SECRET_ACCESS_KEY']]) {
+
+                    // Terminate PGSQL instances (>6 hrs)
+                    // Wrap each call so a single failed instance (already-terminated,
+                    // throttled, wrong region) can't skip the remaining pairs, the stop
+                    // loop below, or the Slack notifications.
+                    pgsqlTerminatePairs.each { pair ->
+                        try {
+                            echo "Terminating PGSQL instance ${pair.id} in region ${pair.region}"
+                            sh "aws ec2 terminate-instances --instance-ids ${pair.id} --region ${pair.region}"
```
nogueiraanderson, line 166: All three mutation loops (163, 174, 202) run from `post always`, so they fire on failure and on abort as well. I aborted a test pipeline mid-loop and the catch swallowed the interrupt, the loop finished, and post still ran. Move the mutations into a stage gated on a successful scan.

misc/ec2-instance-checks.groovy
```diff
@@ -22,125 +22,261 @@ pipeline {
...
+                def buildNumber = currentBuild.number
+                def jobName = env.JOB_NAME
+                def artifactBaseUrl = "${env.JENKINS_URL}/job/${jobName}/${buildNumber}/artifact"
+
+                // === PGSQL: Runs every build (every 6 hrs) ===
+
+                // Parse DESTROY-QA-PGSQL.txt — terminate PGSQL instances running >6 hrs
+                def pgsqlTerminateRegion = ""
+                def pgsqlTerminatePairs = []
+                def terminateLines = readFile('DESTROY-QA-PGSQL.txt').readLines()
```
nogueiraanderson, line 126: Three unguarded `readFile` calls: here, 143, and `OUTPUT-QA.txt` at 188. The one at 188 runs after both reaper loops and its guard only checks whether the general check was meant to run, so a throw there skips every `slackSend` below. Guard each with `fileExists` and move the PGSQL Slack calls ahead of the general block.

misc/ec2-instance-checks.groovy
```diff
@@ -22,125 +22,261 @@ pipeline {
...
-                            env.ovqa = sh(script: "cat ${WORKSPACE}/overview-qa.txt",returnStdout: true).trim()
-                            echo "Print the overview qa after removing the 0 servers list in QA \n ${env.ovqa}"
-                            env.ovqacount=sh(script: """ awk '{sum += \$4} END {print sum}' ${WORKSPACE}/overview-qa.txt """,returnStdout: true).trim()
-             }
 
+
+                        withCredentials([[$class: 'AmazonWebServicesCredentialsBinding', accessKeyVariable: 'AWS_ACCESS_KEY_ID', credentialsId: 'c42456e5-c28d-4962-b32c-b75d161bff27', secretKeyVariable: 'AWS_SECRET_ACCESS_KEY']]) {
+
+                            // Runs every 6 hrs (every build) — PGSQL checks
+                            sh "bash -x ./check-ec2-instances-pgsql.sh"
```
nogueiraanderson, line 54: A failed scan looks identical to a clean estate. With the AWS CLI stubbed to exit 254 the script still exits 0, reports "has 0 INSTANCES" for all nine regions, and the job posts "No PGSQL instances to terminate" on a green build. A preflight `aws sts get-caller-identity` would fail the stage instead. Upstream the script wants `set -euo pipefail`, since `aws ec2 describe-instances | sed` reports sed's status.

misc/ec2-instance-checks.groovy
```diff
@@ -22,125 +22,261 @@ pipeline {
...
                     sudo yum install jq -y
-                    wget https://raw.githubusercontent.com/Percona-QA/package-testing/master/scripts/check-ec2-instances.sh
                     sudo wget -qO /opt/yq https://github.com/mikefarah/yq/releases/latest/download/yq_linux_amd64
                     sudo chmod +x /opt/yq
-                    chmod +x check-ec2-instances.sh
                     """
-             script{
 
+                    script {
+                        env.RUN_EC2_CHECK = (currentBuild.number % 4 == 0).toString()
```
nogueiraanderson, line 34: `currentBuild.number % 4` counts builds, not time. A manual build or replay shifts the phase for good, and a failure on the selected build leaves the general check 48h without running. Two jobs with their own timed lines, 6h and daily, drops the coupling.

misc/ec2-instance-checks.yml
```diff
@@ -21,4 +21,4 @@
         lightweight-checkout: true
         script-path: misc/ec2-instance-checks.groovy
     triggers:
-      - timed: "0 0 * * *"
+      - timed: "0 */6 * * *"
```
nogueiraanderson, line 24: `artifact-num-to-keep` and `num-to-keep` are both 40, so four builds a day turns roughly 40 days of reports into 10. Worth raising both alongside the cron change, since these files are the record of what got terminated.

misc/ec2-instance-checks.groovy
```diff
@@ -22,125 +22,261 @@ pipeline {
...
+
+                        sh """
+                        sed -i '/has 0 INSTANCES WITH PGSQL MOLECULE QA TESTS/d' ${WORKSPACE}/overview-qa-pgsql-to-terminate.txt
+                        sed -i '/has 0 INSTANCES WITH PGSQL MOLECULE QA TESTS/d' ${WORKSPACE}/overview-qa-pgsql-to-stop.txt
+                        """
+
+
+                        env.ovQaPgsqlTerminate = sh(script: "cat ${WORKSPACE}/overview-qa-pgsql-to-terminate.txt", returnStdout: true).trim()
+                        env.ovQaPgsqlStop = sh(script: "cat ${WORKSPACE}/overview-qa-pgsql-to-stop.txt", returnStdout: true).trim()
+                        env.ovQaPgsqlTerminateCount = sh(script: """ awk '{sum += \$4} END {print sum}' ${WORKSPACE}/overview-qa-pgsql-to-terminate.txt """, returnStdout: true).trim()
```
nogueiraanderson, line 87: `ovQaPgsqlTerminateCount` and `ovQaPgsqlStopCount` are never read, the Slack branches use `pairs.size()`.

The regex at 133, 150 and 195 matches `i-` inside a Name tag, so `multi-parallel` parses as `i-parallel`. Use `/\b(i-(?:[0-9a-f]{8}|[0-9a-f]{17}))\b/` and keep the capturing group, since `instanceMatch[0][1]` returns a bare dash without one. Paste this into the Script Console to see all three:

```groovy
def samples = ['  - i-0d71adb8ef92d61bf', '  - ppg-16-component-multi-parallel']
def patterns = [
    'current   ': /(i-[a-zA-Z0-9]+)/,
    'no group  ': /\bi-(?:[0-9a-f]{8}|[0-9a-f]{17})\b/,
    'with group': /\b(i-(?:[0-9a-f]{8}|[0-9a-f]{17}))\b/,
]
samples.each { line ->
    println line.trim()
    patterns.each { name, re -> def m = line =~ re; println "  ${name} -> ${m ? m[0][1] : '(no match)'}" }
}
```

The else branches at 227 and 245 `slackSend` where the general path echoes, up to eight no-op messages a day.

misc/ec2-instance-checks.groovy
nogueiraanderson: Right idea, one bad instance no longer swallows the loop. It needs the failure side in the same commit though: the Slack branch keys off `pgsqlTerminatePairs.size()`, which is IDs parsed rather than terminated, so every call can throw and the job still reports success. Collect the failed IDs into the message and set `currentBuild.result = 'UNSTABLE'`.

The catch swallows aborts too, since `catch (err)` is `catch (Exception)` and `FlowInterruptedException extends InterruptedException`. I aborted a test pipeline mid-loop and the next two iterations ran anyway. Catch `hudson.AbortException` instead, or rethrow the interrupt first.

misc/ec2-instance-checks.groovy
nogueiraanderson: Understood, keeping the behaviour. Ignore my earlier suggestion to narrow it by region, that works against the requirement.

Worth adding an opt-out though. Anything launched with those keys gets stopped every 12h with no exemption, including a long-lived repro box. Skipping instances tagged `no-reap=true` would cover it.

misc/ec2-instance-checks.groovy
nogueiraanderson: Resolved now. The molecule targets carry `jenkins-pg-molecule` and the pg.cd workers are tagged separately, so they sit outside the terminate set. Note that the safety rests on the substring prefix, so any future `jenkins-pg-molecule*` tag joins it automatically.

PR #4336, opened 2026-08-12
PMM-15307 ha tests pipeline

yurkovychv:


pmm/v3/pmm3-ha-rosa.groovy
```diff
@@ -154,6 +154,12 @@ pipeline {
     stages {
         stage('Install CLI Tools') {
             steps {
+                script {
+                    // environment{} block vars are applied via withEnv and are only visible to steps in
+                    // this pipeline - re-assign through env.X= so it's persisted on the build and exposed
+                    // via buildVariables to any caller using build job: 'pmm3-ha-rosa'.
+                    env.CLUSTER_NAME = env.CLUSTER_NAME
```
ademidoff, line 161: That looks like a hack rather than a documented feature. Does it really do anything?

ademidoff (approved):

talhabinrizwan (commented):
**`pmm3-ha-eks.groovy` needs the same fix**

`pmm3-ha-eks.groovy:66` has the same `environment{}` issue — so on the EKS path `buildVariables.CLUSTER_NAME` is `null`, the `post{always}` guard fails, and `pmm3-ha-eks-cleanup` never runs.

pmm/v3/pmm3-ha-rosa.groovy
yurkovychv: This export allows me to capture `env.CLUSTER_NAME` in the `clusterCreateJob.buildVariables` as without this export it is just null

talhabinrizwan (approved):

PR #4348, opened 2026-08-17
Auto-update OL/Rocky molecule AMI IDs weekly instead of resolving at test time

Naeem-Akhter:
Adds an update-molecule-env job that runs after a successful weekly bake and rewrites vars/moleculeEnvPPG.groovy with the newest AMI per os+major+arch, replacing the dynamic DescribeImages call every parallel molecule job used to make independently (source of the RequestLimitExceeded errors) with a single resolution done once here, matching how RHEL/Ubuntu/Debian are already hand-pinned.

### DIFF
--- a/.github/workflows/ppg-ami-factory.yml
+++ b/.github/workflows/ppg-ami-factory.yml
@@ -180,6 +180,48 @@ jobs:
           done
           echo "- promote: $pr" >> "$GITHUB_STEP_SUMMARY"
 
+  update-molecule-env:
+    name: Update moleculeEnvPPG.groovy (OL/Rocky AMI IDs)
+    needs: bake
+    # Only after a real weekly rebake — never on manual/PR/push runs, where a
+    # partial or test-env combo list must not overwrite the pinned AMI IDs.
+    if: success() && github.event_name == 'schedule'
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    permissions:
+      contents: write
+      id-token: write
+    steps:
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
+        with:
+          ref: master
+
+      - name: Configure AWS credentials (OIDC)
+        uses: aws-actions/configure-aws-credentials@7474bc4690e29a8392af63c5b98e7449536d5c3a # v4.3.1
+        with:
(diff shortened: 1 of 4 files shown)


nogueiraanderson:
Tested this end to end and pushed fixes for everything found, so the PR is ready for review as it stands ([commits](https://github.com/Percona-Lab/jenkins-pipelines/pull/4348/commits)).

What I verified before changing anything: the script resolves all 12 combos against live AWS, the rewrite of `vars/moleculeEnvPPG.groovy` is correct and idempotent, and the result is valid Groovy and shell. The diagnosis is right too: 30 OS stages x 12 `DescribeImages` calls = 360 per build, which trips the account-wide EC2 rate limit.

What was broken in the original approach, each verified empirically and now fixed:

1. **Direct push to master is rejected** (`GH006`, reproduced in a [probe repo](https://github.com/nogueiraanderson/probe-pr4348-gha/actions/runs/32268633114) with the same protection settings). The weekly refresh now publishes a rolling `ppg-ami-refresh` branch and auto-opens a PR.
2. **Updater failures were Slack-silent**: `notify` only watched `bake`. Now `needs: [bake, update-molecule-env]`.
3. **`if: success()` skipped the refresh on any partial bake**, exactly what the [2026-08-17 weekly run](https://github.com/Percona-Lab/jenkins-pipelines/actions/runs/32001785828) did (10 of 12 combos green). Now `!cancelled()`, a failed combo keeps last week's promoted AMI.
4. **`boto3`/`botocore` pinned**, the job holds a write token and AWS credentials.
5. **`vars/moleculeEnvPPG.groovy` converted to static pins in this PR** (generated by the PR's own script from the [original dynamic block](https://github.com/Percona-Lab/jenkins-pipelines/blob/0315b1550e434592afd48befc44a8702fb200356/vars/moleculeEnvPPG.groovy#L6-L37)), so the throttle stops at merge instead of waiting for the first green Monday.

One item to confirm on the first weekly run: PR creation needs the repo setting "Allow GitHub Actions to create and approve pull requests". If it is off, the job fails loudly and the new notify wiring alerts.

Tracked in [PKG-1447](https://perconadev.atlassian.net/browse/PKG-1447).


[PKG-1447]: https://perconadev.atlassian.net/browse/PKG-1447?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

EvgeniyPatlan (approved):

PR #4350, opened 2026-08-18
PMM-15341 Publish a self-contained kubeconfig from the PMM HA cluster jobs

talhabinrizwan:
https://perconadev.atlassian.net/browse/PMM-15341

### DIFF
--- a/pmm/v3/pmm3-ha-rosa.groovy
+++ b/pmm/v3/pmm3-ha-rosa.groovy
@@ -1,3 +1,56 @@
+// Writes ${WORKSPACE}/kubeconfig with a ServiceAccount token inside it, replacing the OAuth token
+// from "oc login" that expires after 24h. This one never expires, so it lasts as long as the cluster.
+def mintAdminKubeconfig() {
+    sh '''
+        set +x   # Jenkins runs sh with -xe; keep the token out of the console
+
+        MINT_OUT="${WORKSPACE}/kubeconfig"
+
+        kubectl -n kube-system create serviceaccount pmm-ha-admin
+        kubectl create clusterrolebinding pmm-ha-admin \
+            --clusterrole=cluster-admin \
+            --serviceaccount=kube-system:pmm-ha-admin
+
+        # Unlike "kubectl create token", the token in this Secret never expires.
+        kubectl apply -f - <<'EOF'
+apiVersion: v1
+kind: Secret
+metadata:
+  name: pmm-ha-admin-token
+  namespace: kube-system
+  annotations:
+    kubernetes.io/service-account.name: pmm-ha-admin
+type: kubernetes.io/service-account-token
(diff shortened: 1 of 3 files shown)


pmm/v3/pmm3-ha-rosa.groovy
```diff
@@ -1,3 +1,56 @@
+// Writes ${WORKSPACE}/kubeconfig with a ServiceAccount token inside it, replacing the OAuth token
+// from "oc login" that expires after 24h. This one never expires, so it lasts as long as the cluster.
+def mintAdminKubeconfig() {
```
yurkovychv, line 3: This function seems to be indentical for both files

talhabinrizwan: Not quite identical, the final check differs, on EKS it runs with the AWS credentials stripped, since a leftover exec stanza would otherwise pass using the pipeline's own credentials.

The duplication is deliberate. I tested all of this by replaying with an edited script, and an in-file def is editable in Replay where a library function isn't. Neither pipeline imports a shared library today either, so `vars/` would add that dependency to both.

pmm/v3/pmm3-ha-rosa.groovy
```diff
@@ -7,24 +60,15 @@ def cleanupCluster() {
...
-            # Capture cluster ID before deletion (cluster may already be partly gone if create failed midway)
-            CLUSTER_ID=""
-            if [ -f ${HOME}/cluster-id.txt ]; then
-                CLUSTER_ID=$(cat ${HOME}/cluster-id.txt)
-            elif rosa describe cluster --cluster="${CLUSTER_NAME}" --region="${REGION}" >/dev/null 2>&1; then
-                CLUSTER_ID=$(rosa describe cluster --cluster="${CLUSTER_NAME}" --region="${REGION}" -o json | jq -r '.id // empty')
-            fi
+            # Read before the delete below, while the cluster still exists. Empty if it never did.
+            CLUSTER_ID=$(rosa describe cluster --cluster="${CLUSTER_NAME}" --region="${REGION}" \
+                -o json 2>/dev/null | jq -r '.id // empty')
```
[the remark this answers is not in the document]
talhabinrizwan: `rosa describe cluster` does work for error and uninstalling clusters, those are still in OCM (OpenShift Cluster Manager). It only fails once the record is gone, and in that case `cluster-id.txt` wouldn't exist either, it was written after rosa create cluster returned. So the file was redundant with the live lookup rather than a fallback for it.

The read now happens before the delete, so the ID survives into the sweep after the cluster is gone from OCM, which is the case the sweep is for. I've tested both aborting mid-creation and aborting during the ready-wait, cleanup behaved correctly in each.

ademidoff (approved):

PR #4358, opened 2026-08-19
Mysql shell build matrix

EvgeniyPatlan:


adivinho:
Hi @EvgeniyPatlan Could you check why the job was never succesful?
Screenshot 2026-08-20 at 13 51 30

PR #4383, opened 2026-08-28
PMM: build and release pmm-server for arm64

EvgeniyPatlan:
pmm3-server-autobuild becomes a parent that runs two new per-arch jobs (pmm3-server-autobuild-amd, pmm3-server-autobuild-arm) in parallel and merges the images into a multi-arch manifest, same layout as the client autobuild. API tests, image scanning, staging and package testing get arch parameters.

The release job promoted the server with docker tag/push, which flattens the manifest to the arch of the agent doing the push, so it now uses buildx imagetools the way the client already does. Server docker tarballs are published per arch, the arch-less name stays as the amd64 one so the existing download links keep working.

The AMI jobs take AMI_ARCH and build with the pmm-ami-arm64 packer target, staging picks Graviton instance types and installs a matching docker-compose. The RC builds both AMIs and its testing lane runs the docker and AMI e2e suites against the arm server, the arm package matrix runs against an arm server too instead of an amd64 one.

Both client autobuilds uploaded pmm-client-<version>.tar.gz to the same directory in TESTING/pmm, so the arm tarball could overwrite the amd64 one. They carry the arch suffix now, x86_64 keeps the old name as well.


pmm/v3/pmm3-package-testing-arm64.groovy
```diff
@@ -115,6 +116,10 @@ pipeline {
             description: 'PMM Server docker container version (image-name:version-tag)',
             name: 'DOCKER_VERSION',
             trim: true)
+        choice(
+            choices: ['arm64', 'amd64'],
```
nogueiraanderson: First choice is the default, so existing callers silently move their staging server to arm64. Swap to `['amd64', 'arm64']` and opt into arm64 explicitly (same in `pmm3-package-tests-matrix-arm64.groovy`).

pmm/v3/pmm3-server-autobuild-amd.groovy
```diff
@@ -0,0 +1,180 @@
...
+                    set -o errexit
+                    git submodule update --init --jobs 10
+                    git submodule status
+
+                    git rev-parse --short HEAD > shortCommit
+                    echo "UPLOAD/pmm3-components/yum/${DESTINATION}/${JOB_NAME}/pmm/${VERSION}/${GIT_BRANCH}/$(cat shortCommit)/${BUILD_NUMBER}" > uploadPath
+                '''
+                script {
+                    if (params.DESTINATION == "testing") {
+                        env.DOCKER_LATEST_TAG     = "${VERSION}-rc${BUILD_NUMBER}-amd64"
```
nogueiraanderson: The amd and arm children use their own build numbers here, so the two `rc<N>-<arch>` tags never pair and the parent never merges them, and RC testing lands on the mutable `${VERSION}-rc` tag. Merge a `${VERSION}-rc${BUILD_NUMBER}` multi-arch tag in the parent instead and drop these per-child rcN pushes.

pmm/v3/vars/runSpotInstance.groovy
```diff
@@ -8,7 +8,7 @@ def call(String INSTANCE_TYPE, boolean USE_ONDEMAND = false) {
             IMAGE_ID=$(
                 aws ec2 describe-images \
                     --owners self \
-                    --filters "Name=tag:iit-billing-tag,Values=pmm-worker-3" "Name=architecture,Values=x86_64" \
+                    --filters "Name=tag:iit-billing-tag,Values=pmm-worker-3" "Name=architecture,Values=${ARCH}" \
                     --region us-east-2 \
                     --output text \
                     --query 'Images[0].ImageId'
```
nogueiraanderson, line 14: A no-match lookup returns the literal `None` and the spot loop below (lines 57 to 117) then retries forever. Add `[ "$IMAGE_ID" = "None" ] && exit 1` here and cap the loop attempts.

pmm/v3/pmm3-server-autobuild.groovy
```diff
@@ -25,140 +24,92 @@ pipeline {
     }
     options {
```
nogueiraanderson, line 25: Without `skipDefaultCheckout()` the pmm-submodules checkout at line 36 overlays the jenkins-pipelines clone. Restore `skipDefaultCheckout()` or add the client sibling's `git reset --hard && git clean -xdf`.

nogueiraanderson (changes requested):
Thanks, the fan-out design is sound and the July canary proved both arch builds and the manifest merge work. Four blockers:

1. `pmm3-server-autobuild-amd` and `pmm3-server-autobuild-arm` do not exist on the pmm controller. Create both jobs and run each once before merge, otherwise the next `pmm3-submodules-rewind` trigger fails the parent.
2. The release drops the documented `pmm-server-<version>.docker` and `.sha256sum`. The isolated-hosts install guide still downloads them. Keep publishing the archless amd64 copy, like the client repo-push already does.
3. Image scanning fails open. `perconalab/pmm-server:3-dev-latest` is amd64-only today and half-scans green, and `percona/pmm-server:2` scans nothing and stays green. Parse platforms with jq, fail on a missing arch, and drop `allowEmptyArchive`.
4. `perconalab/watchtower:dev-latest` has no arm64 build, so every arm64 staging start fails at `docker run`. Publish a multi-arch tag or skip watchtower when `SERVER_ARCH=arm64`.

Smaller items inline. An arm64 staging AMI already exists in us-east-2.

pmm/v3/pmm3-aws-staging-start.groovy
```diff
@@ -23,6 +23,11 @@ pipeline {
             description: 'WatchTower docker container version (image-name:version-tag, ex: perconalab/watchtower:dev-latest)',
             name: 'WATCHTOWER_VERSION'
         )
+        choice(
+            choices: ['amd64', 'arm64'],
+            description: 'CPU architecture of the staging VM (arm64 = AWS Graviton t4g.xlarge)',
+            name: 'SERVER_ARCH'
+        )
```
talhabinrizwan, line 30: `SERVER_ARCH=arm64` can't start today: both `perconalab/watchtower:dev-latest` and `percona/watchtower:latest` are amd64-only (no arm64 in either manifest), so the `docker run --name watchtower` at line 245 fails under `set -o errexit` and the staging start dies there. That takes out the new arm64 defaults in `pmm3-package-testing-arm64` / `pmm3-package-tests-matrix-arm64` and the arm64 lanes in `pmm3-rc-testing`.

We want to keep watchtower for now to test pre-3.9.0 upgrades, and no pre-3.9.0 `pmm-server` image is published for arm64, so gating on the arch is enough. Couldn't attach this as an applicable suggestion because the block isn't in the diff — patch for lines 245-253:

```groovy
                                    # watchtower has no arm64 image, and only PMM before 3.9.0 (amd64 only) needs it
                                    if [ "${SERVER_ARCH}" != "arm64" ]; then
                                        docker run --detach --restart always \
                                            --network pmm-qa \
                                            --name watchtower \
                                            --volume /var/run/docker.sock:/var/run/docker.sock \
                                            --restart always \
                                            -e WATCHTOWER_DEBUG=1 \
                                            -e WATCHTOWER_HTTP_API_TOKEN=testToken \
                                            -e WATCHTOWER_HTTP_API_UPDATE=1 \
                                            ${WATCHTOWER_VERSION}
                                    fi
```

`-e PMM_WATCHTOWER_HOST` / `-e PMM_WATCHTOWER_TOKEN` on the `pmm-server` container can stay as they are — arm64 only ever runs 3.9.0+, which ignores them.

talhabinrizwan (approved):
LGTM

peterSirotnak (approved):

PR #4393, opened 2026-08-31
CLOUD-934 Refactor pxco and psmdbo release jobs

jvpasinatto:
- Refactor PXC and PSMDB operator release jobs to use common functions to simplify the code
- Add error for platforms that don't support different archs


cloud/common/vars/tests.groovy
```diff
@@ -249,10 +364,17 @@ void resolveMachineType(Map testVariables) {
 }
 
 Map prepareVersions(Map testVariables) {
+    validateTestVariables(testVariables)
```
valmiranogueira, line 367: I think that clusters and number of clusters must be enforced in pipeline side not libraries

cloud/jenkins/psmdbo_eks.groovy
```diff
@@ -274,83 +42,131 @@ pipeline {
...
+        }
+
         stage('Prepare Node') {
             steps {
-                prepareNode()
+                script {
+                    libraries.dependencies.prepareNode(
+                        libraries,
+                        'make',
+                        'psmdb-operator',
```
valmiranogueira: Not now perhaps in future we can have a standard between psmdb or psmdb-operator (I think only psmdb would be better :D)

cloud/jenkins/psmdbo_aks.groovy
```diff
@@ -277,83 +50,131 @@ pipeline {
...
+
         stage('Init Tests') {
             steps {
-                initTests()
+                script {
+                    libraries.tests.initTestRun(testVariables, [
+                        testList          : TEST_LIST,
+                        testSuite         : TEST_SUITE,
+                        ignorePreviousRun : IGNORE_PREVIOUS_RUN,
+                        operator          : 'psmdb'
```
valmiranogueira, line 153: Previous comment was about this 'psmdb'

eleo007 (approved):

cloud/jenkins/psmdbo_eks.groovy
jvpasinatto: done

cloud/common/vars/tests.groovy
jvpasinatto: why specifically ? we would have to repeat the same function in every job.

PR #4430, opened 2026-09-08
Install upgrade-source clients the apt repo no longer carries from tarballs

travagliad:

- source: https://pmm.cd.percona.com/job/pmm3-nightly-orchestrator/3/ — `pmm3-upgrade-test-runner` [#23452](https://pmm.cd.percona.com/job/pmm3-upgrade-test-runner/23452/) (3.7.0 SSL), [#23453](https://pmm.cd.percona.com/job/pmm3-upgrade-test-runner/23453/) (3.7.0 EXTERNAL SERVICES), [#23454](https://pmm.cd.percona.com/job/pmm3-upgrade-test-runner/23454/) (3.7.0 OTHERS)
- tests: none — all three died in `Setup Databases for PMM-Server`, so no test ever ran.
- companion PR: percona/pmm-qa#1371 (required — see "Why two PRs")

## What actually failed

Not a test, not infra, and not the fan-out's load. The `pmm-client 3.7.0` install inside the setup containers 404s:

```
+ wget -O pmm-client.deb https://repo.percona.com/pmm3-client/apt/pool/main/p/pmm-client/pmm-client_3.7.0-7.jammy_amd64.deb
HTTP request sent, awaiting response... 404 Not Found
```

`pmm-client 3.7.0` is in **no** component of the pmm3-client apt repo — I checked `main`, `testing`, `experimental` and `laboratory` for jammy, and `main` for noble/focal/bookworm. The flat pool carries only `3.0.0`, `3.1.0`, `3.2.0` (focal-only) and `3.7.1`, `3.8.0`, `3.8.1`, `3.9.0`, `3.9.1`. No `-<revision>` guess can fix it: `-7`, `-6` and `-1` all 404 individually.

3.7.0 is a real release, though — `percona/pmm-server:3.7.0` and `percona/pmm-client:3.7.0` both exist, and its **client tarball is published and installable**. So this is not a case for dropping 3.7.0 from `upgradeBranches`.

## Root cause

`upgradeBranches` already had a tarball fallback for exactly this. It could never fire:

```groovy
def upgradeVersions = pmmVersion('v3')[-6..-1]   // 3.7.0 … 3.9.1
def oldVersions     = pmmVersion('v3-old')       // v3[0..-7] → 3.0.0 … 3.6.0
…
: (ver in oldVersions ? "<tarball URL>" : ver)
```

`v3-old` is `v3[0..-7]` and the matrix iterates `v3[-6..-1]` — **exact complements**, so `ver in oldVersions` is false for every version iterated and the tarball branch is dead code. Every source therefore installed from the hand-built pool URL.

That collides with the repo's retention: the apt pool keeps the newest **five** client debs, the matrix upgrades from the newest **six**. The oldest of the six always has no deb. This is structural and recurring, not a 3.7.0 quirk — 3.7.1 is next when 3.10.0 ships.

`pmmVersion` itself is correct ("old" genuinely means older than the last six) and lives in the repo-root shared `vars/`, so it is untouched; only the wrong consumer changed. **This bug is not new to the orchestrator** — `pmm/v3/pmm3-upgrade-tests-matrix.groovy` has the identical complementary-lists mistake and will fail the same way for the same reason.

## The change

Ask the apt index what it actually carries instead of assuming a retention depth, and tarball anything missing from it. If the index can't be read, the list stays empty and every source uses its tarball, rather than failing the whole 50-suite nightly at planning time.

With today's index (`3.7.1 3.8.0 3.8.1 3.9.0 3.9.1`) the selection is:

| source | before | after |
|---|---|---|
| 3.7.0 | deb (404) | tarball |
| 3.7.1 / 3.8.0 / 3.8.1 / 3.9.0 | deb | deb (unchanged) |
| 3.9.1 (newest) | `pmm3-rc` | `pmm3-rc` (unchanged) |

Only the broken variant changes behaviour; the four passing deb variants are deliberately left on the deb path.

## Verified

On a throwaway Linode VM against `percona/pmm-server:3.7.0`, pmm-qa at `c4ae7ab`:

- **Reproduced** the CI failure exactly — `pmm3-client-setup.sh --client_version 3.7.0` exits 127 on `ERROR 404`, from a single request on an idle box. Rules out the fan-out's ~24 GB burst, and rules out percona/pmm-qa#1366 (a `--continue`/timeout hardening cannot fix a 404).
- **Verified the fix** — the same script with the 3.7.0 tarball URL exits 0, installs `pmm-admin`/`pmm-agent` 3.7.0 and registers: `Connected: true`, `pmm-admin version: 3.7.0`.
- **Verified the failing stage** — `pmm-framework --database ssl_pdpgsql --client-version <3.7.0 tarball>` finishes `failed=0`, "playbook execution successful", against `failed=1` in CI.
- Tarball URLs return 200 for every version the matrix can reach (3.5.0 → 3.9.1), so the fallback is not 3.7.0-specific.

**Not verified:** this Groovy has not been executed. No Jenkins run and no Groovy parser was available to me, so the pipeline change is reviewed-by-reading — the selection logic and the `awk` probe were validated against the real apt index outside Jenkins. The next orchestrator run is what confirms it; the `client source :` line added to the Plan echo makes the choice visible without digging into a child job.

Also worth a maintainer's eye: `ssl_mysql` failed identically in #23452 and is not in the run's stage summary, so the same 404 hits every setup leg, not just `ssl_pdpgsql`.

## Why two PRs

Enabling the tarball route alone just moves the failure. `check_client_upgrade.py` compares `pmm-admin`'s version against `CLIENT_VERSION` verbatim, so a URL can never match and `Check Client before Upgrade` fails with a correctly installed client — confirmed on the box (exit 1 with the URL, exit 0 with `3.7.0`, error text showing the client was in fact 3.7.0). percona/pmm-qa#1371 fixes that and must land with this. Because that route has been dead since it was written, it had never been exercised.

### DIFF
--- a/pmm/v3/pmm3-nightly-orchestrator.groovy
+++ b/pmm/v3/pmm3-nightly-orchestrator.groovy
@@ -295,18 +295,16 @@ timestamps {
                     returnStdout: true,
                     script: 'curl -fsSL https://raw.githubusercontent.com/Percona-Lab/pmm-submodules/v3/VERSION'
                 ).trim()
-                // Ask the apt index which client debs still exist rather than
-                // assuming a retention depth; the client containers are jammy.
-                // A repo blip leaves the list empty, which installs every source
-                // from its tarball rather than failing the whole nightly here.
-                try {
-                    clientDebVersions = sh(
-                        returnStdout: true,
-                        script: '''curl -fsSL https://repo.percona.com/pmm3-client/apt/dists/jammy/main/binary-amd64/Packages | awk '/^Version:/{split($2,a,"-"); print a[1]}' | sort -u'''
-                    ).trim().tokenize()
-                } catch (err) {
-                    echo "Could not read the pmm3-client apt index (${err.message}); every upgrade source will install from its tarball."
-                }
+                // Ask the apt index which client debs exist rather than assuming a
+                // retention depth. jammy, noble and bookworm carry identical
+                // versions and revisions, so one query covers them; focal does not
+                // (3.0.0-3.2.0 only), so this is a proxy for the client containers
+                // that are focal. Resolving per container -- where lsb_release and
+                // the arch are actually known -- is the fix that removes the proxy.
+                clientDebVersions = sh(
+                    returnStdout: true,
+                    script: '''curl -fsSL https://repo.percona.com/pmm3-client/apt/dists/jammy/main/binary-amd64/Packages | awk '/^Package: pmm-client$/{p=1} p&&/^Version:/{split($2,a,"-"); print a[1]; p=0}' | sort -u'''
+                ).trim().tokenize()
             } finally {
                 deleteDir()
             }
--- a/pmm/v3/pmm3-nightly-orchestrator.groovy
+++ b/pmm/v3/pmm3-nightly-orchestrator.groovy
@@ -236,7 +236,7 @@ def packageBranches(Map branches, String prefix, String jobName, String serverAr
     }
 }
 
-def upgradeBranches(Map branches, List pmmVersions, List oldVersions, String latestVersion, String latestDevVersion) {
+def upgradeBranches(Map branches, List pmmVersions, List clientDebVersions, String latestVersion, String latestDevVersion) {
     // Mirrors pmm3-upgrade-tests-matrix: every recent version upgraded, with the
     // newest one going from its RC image up to the dev tip.
     def variants = ['SSL', 'EXTERNAL SERVICES', 'OTHERS']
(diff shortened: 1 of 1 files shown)


pmm/v3/pmm3-nightly-orchestrator.groovy
```diff
@@ -236,16 +236,20 @@ def packageBranches(Map branches, String prefix, String jobName, String serverAr
...
     // newest one going from its RC image up to the dev tip.
     def variants = ['SSL', 'EXTERNAL SERVICES', 'OTHERS']
     pmmVersions.each { ver ->
         def isNewest = (ver == pmmVersions.last())
         def dockerTag = isNewest ? "perconalab/pmm-server:${ver}-rc" : "percona/pmm-server:${ver}"
         def dockerTagUpgrade = isNewest ? 'perconalab/pmm-server:3-dev-latest' : "perconalab/pmm-server:${pmmVersions.last()}-rc"
+        // repo.percona.com keeps only the newest few client debs, so the oldest
+        // sources in this matrix have no deb to install and pmm3-client-setup.sh's
+        // constructed pool URL 404s. clientDebVersions is what the apt index
+        // actually offers; anything missing from it installs from its tarball.
```
```suggestion
```

travagliad: Dropped in `0af2a049`, which also resolves the conflict #4428 created by landing on the base first.

The conflict was one hunk and it was keep-both, as this PR's body predicted: `packageBranches` takes #4428's new `gaVersion` argument, and `upgradeBranches` keeps `clientDebVersions` in place of the dead `oldVersions` three lines below. Merged rather than rebased, so anyone with this branch checked out keeps a valid checkout.

Checked after resolving, because a keep-both on adjacent lines is exactly where a resolution quietly loses one side:

- no conflict markers, and no stale `oldVersions` reference anywhere — its definition went with this branch's change, so a leftover call would have been an undefined variable at Plan time
- `clientDebVersions` defined once and referenced at all three sites (the probe, the `upgradeBranches` call, and the Plan echo)
- both halves present: `gaVersion`/`gaOnly` and the `(GA …)` lane naming from #4428, the apt-index probe from here
- `npm-groovy-lint`: **0 errors** (22 warnings, all pre-existing `LineLength`)

---
_Generated by [Claude Code](https://claude.ai/code)_

pmm/v3/pmm3-nightly-orchestrator.groovy
```diff
@@ -295,6 +295,16 @@ timestamps {
                     returnStdout: true,
                     script: 'curl -fsSL https://raw.githubusercontent.com/Percona-Lab/pmm-submodules/v3/VERSION'
                 ).trim()
+                // Ask the apt index which client debs exist rather than assuming a
+                // retention depth. jammy, noble and bookworm carry identical
+                // versions and revisions, so one query covers them; focal does not
+                // (3.0.0-3.2.0 only), so this is a proxy for the client containers
+                // that are focal. Resolving per container -- where lsb_release and
+                // the arch are actually known -- is the fix that removes the proxy.
```
```suggestion
```

PR #4440, opened 2026-09-11
Fix build pipeline selection for pxb-new-ps-* jobs

tplavcic:
When PXB is built from branch if 8.4 was selected it was using 8.0-compile-pipeline job instead of 8.1-compile-pipeline.


kaushikpuneet07 (approved):
LGTM

PR #4441, opened 2026-09-11
PMM-13901 Fix arm64 spot capacity, PMM client install and spot instance cleanup.

talhabinrizwan:
`t4g.xlarge` spot capacity in `us-east-2` is thin (AWS placement score 1/10 per arm64 type, in every AZ), so the 5-attempt loop exhausted before a VM launched. Fixing that surfaced several further defects.

### `vars/runSpotInstance.groovy` — rebuilt around `run-instances`

Replaces `request-spot-instances` with a single synchronous `aws ec2 run-instances --instance-market-options MarketType=spot`, looping over every `pmm-staging` subnet × each candidate instance type and stopping at the first success. 173 lines → 106.

The old flow created a spot *request*, polled ~10s for a public IP, then cancelled and moved on. That caused four separate problems, which this removes rather than patches:

- **Capacity.** One randomly drawn subnet, one instance type. Now 3 AZs × 3 arm64 types (`t4g`/`m6g`/`m7g.xlarge` — all 4 vCPU / 16 GiB, so a fallback never downgrades the VM). Non-arm64 stays a single type.
- **The bid was filtering out the fallbacks.** `SpotPrice` was derived from `t4g.xlarge` alone, then bid 1.15×. m6g and m7g cost more, so those requests sat `price-too-low` until round 3. Dropping `SpotPrice` caps the bid at on-demand — we still pay market, we just stop being refused. (@ademidoff, @nogueiraanderson)
- **A cancelled request could leave a VM running.** Cancelling doesn't terminate an instance the request already launched, and the job had neither its id nor any tags on it. `run-instances` is synchronous: no request outlives the call, so there is nothing to cancel.
- **The untagged window.** Tags now apply at launch via `--tag-specifications`, so a VM is never running while invisible to `aws-staging-stop`.

Failing fast is also quicker: `run-instances` answers in ~1.9s, so a full 9-combination sweep takes under 20s, against the 150s the old loop spent on effectively one combination.

### `pmm3-aws-staging-start.groovy`

- Pass `CLIENT_VERSION` to `pmm-framework` verbatim. The job rewrote `3-dev-latest → latest`, which matches no branch in `install_pmm_client.yml`, so every install task skipped and `pmm-agent` setup failed with rc 127. Arch-independent, broke when `pmm-qa #1125` inverted the flag/env precedence on 2026-07-31, and went unnoticed because every green build since had `CLIENTS` empty or `--help`.
- Cleanup moved from `failure` to `unsuccessful` so aborted builds clean up too, and it reads the workspace files rather than the env vars, which are only set once `runSpotInstance` returns. Both `cat` reads are `|| true`-guarded — Jenkins runs the step as `sh -xe`, so without it a missing file aborted the block before it cleaned anything. (@nogueiraanderson)
- Strip whitespace from `--client-version`: quoting it preserved a trailing space the old unquoted expansion dropped. (@nogueiraanderson)
- Drop `pmm3-latest` from the `CLIENT_VERSION` description — nothing in the repo passes it, and on the host it is byte-identical to `3-dev-latest`. Note that `latest-tarball` gives the docker clients an amd64 tarball.

### `pmm3-deploy-services.groovy`

Same description defect at line 123: it advertised `"latest"` and passes `CLIENT_VERSION` straight into `pmm3-aws-staging-start`, so it was a second door to the bug above.

### Testing

| build | covers | result |
|---|---|---|
| 25649 | arm64 + spot | SUCCESS |
| 25651 | amd64 + spot | SUCCESS |
| 25653 | arm64 + on-demand | SUCCESS |
| 25655 | amd64 + on-demand | SUCCESS |
| 25656 | fallback: 5 injected failures, then launch | SUCCESS |
| 25657 | exhaustion: every combination fails | FAILURE (expected) |

25656 and 25657 were fault-injected via Replay (a bogus subnet id, and `t3.xlarge` against the arm64 AMI). 25656 confirms a failing `run-instances` inside the `if` does not trip `errexit`. 25657 confirms the exhaustion message, `exit 1`, no leaked instance, and that `post { unsuccessful }` runs to completion with both files empty.

Known follow-up, outside this PR: `latest-tarball` and `X.X.X` hand docker clients an amd64 client on arm64 VMs — `normalize_client_version` (`pmm-framework/lib/common.sh:78`) and `install_pmm_client.yml:157,174` hardcode amd64. Will fix in pmm-qa.


pmm/v3/vars/runSpotInstance.groovy
```diff
@@ -87,41 +92,43 @@ def call(String INSTANCE_TYPE, boolean USE_ONDEMAND = false, String ARCH = 'x86_
                             "SubnetId": "$SUBNET"
                         },
                         "SpotPrice": "$SPOT_PRICE",
-                        "Type": "persistent"
+                        "Type": "one-time"
```
ademidoff, line 95: This would be the game changer. Really sure we want to do it?

pmm/v3/pmm3-aws-staging-start.groovy
```diff
@@ -364,17 +361,21 @@ pipeline {
                 }
             }
         }
-        failure {
+        unsuccessful {
```
ademidoff, line 364: What is the difference between `failure` and `unsuccessful`? We never use the latter. Therefore, for consistency sake here is the question: are we migrating?

pmm/v3/vars/runSpotInstance.groovy
```diff
@@ -58,20 +62,21 @@ def call(String INSTANCE_TYPE, boolean USE_ONDEMAND = false, String ARCH = 'x86_
...
+                        cat > config.json <<EOF
                       {
                         "DryRun": false,
                         "InstanceCount": 1,
                         "InstanceInterruptionBehavior": "terminate",
                         "LaunchSpecification": {
                             "EbsOptimized": false,
                             "ImageId": "$IMAGE_ID",
-                            "InstanceType": "$INSTANCE_TYPE",
+                            "InstanceType": "$TYPE",
```
ademidoff, line 79: `$TYPE` now varies, but `$SPOT_PRICE` is still derived from `$INSTANCE_TYPE` only (line 57 — t4g.xlarge). m6g/m7g spot sits above t4g, so at the low multipliers those two requests bid below market, stay `open`/`price-too-low`, and just burn their 10s before being cancelled.

Either query the price for all of `$CANDIDATE_TYPES` and take the max, or move the lookup inside the `for TYPE` loop.

pmm/v3/vars/runSpotInstance.groovy
```diff
@@ -87,41 +92,43 @@ def call(String INSTANCE_TYPE, boolean USE_ONDEMAND = false, String ARCH = 'x86_
                             "SubnetId": "$SUBNET"
```
ademidoff, line 92: `$SUBNET` is drawn once before the loop, so all 15 attempts are pinned to a single AZ. m6g/m7g.xlarge aren't offered in every us-east-2 AZ — one unlucky draw and every fallback request comes back `constraint-not-fulfillable`, i.e. the fan-out is a no-op. Placement scores are region-level; per-AZ they can be 1/10.

Re-roll `SUBNET` per attempt, or request across all pmm-staging subnets.

pmm/v3/vars/runSpotInstance.groovy
```diff
@@ -87,41 +92,43 @@ def call(String INSTANCE_TYPE, boolean USE_ONDEMAND = false, String ARCH = 'x86_
...
 
-                    REQUEST_ID=$(
-                        aws ec2 request-spot-instances \
-                            --output text \
-                            --region us-east-2 \
-                            --cli-input-json file://config.json \
-                            --query 'SpotInstanceRequests[].SpotInstanceRequestId' \
-                            | tee REQUEST_ID
-                    )
+                        REQUEST_ID=$(
```
ademidoff, line 99: `$REQUEST_ID` is never checked for emptiness and this `sh` sets only `xtrace`, no `errexit`. If `request-spot-instances` fails for a candidate (type not offered in the drawn AZ, or a throttle from the new 3x call volume), the id is empty and execution continues: line 112 filters on `Values=` and line 124 calls `cancel-spot-instance-requests` with no argument. Both errors just scroll by and the build reports "Could not get a spot instance" — a real API/config error looks identical to no capacity. The empty `REQUEST_ID` file is also what the new `post { unsuccessful }` reads.

`[ -z "$REQUEST_ID" ] && continue`.

pmm/v3/vars/runSpotInstance.groovy
```diff
@@ -87,41 +92,43 @@ def call(String INSTANCE_TYPE, boolean USE_ONDEMAND = false, String ARCH = 'x86_
...
+                        done
 
-                    if [ -s IP ]; then
-                        break
-                    fi
+                        if [ -s IP ]; then
+                            break
+                        fi
+
+                        aws ec2 cancel-spot-instance-requests --region us-east-2 --spot-instance-request-ids $REQUEST_ID
```
ademidoff, line 124: Cancelling a request doesn't terminate an instance that already launched. If fulfilment lands just after `ATTEMPTS` hits 0, this leaks a running instance: `create-tags` is after the loop so it has no `iit-billing-tag`/`stop-after-days` (invisible to `pmm/aws-staging-stop.groovy`), and `AMI_ID` isn't written yet so `post { unsuccessful }` can't terminate it either. Previously once per price level, now up to 15x per build.

Worth a `describe-instances --filters Name=spot-instance-request-id,Values=$REQUEST_ID` + terminate before moving to the next type.

pmm/v3/vars/runSpotInstance.groovy
ademidoff: Right now the job offers to keep for up to 30 days. This basically makes the `days` parameter useless, with no compensating signal to the owner.

ademidoff (changes requested):

pmm/v3/pmm3-aws-staging-start.groovy
talhabinrizwan: `unsuccessful` fires on anything that isn't `SUCCESS` i.e. `FAILURE`, `ABORTED`, and `UNSTABLE`. I used `unsuccessful` rather than adding `aborted { }` to avoid duplicating the cleanup in two post blocks.

nogueiraanderson (commented):
MERGE-AFTER-FIXES. Two things block: the post block exits under errexit before it cancels anything (line 371), and `pmm3-latest` is advertised as the released client while the host installer pulls experimental (line 33). The cancel race is bounded by the hourly stop robot, and the pricing and subnet issues cost rounds rather than break the job. The CLIENT_VERSION change and `unsuccessful` with file reads are right. I can follow up with a `run-instances --instance-market-options MarketType=spot` version that answers synchronously and tags at launch, tested live against the pmm-staging subnets today.

pmm/v3/pmm3-aws-staging-start.groovy
```diff
@@ -364,17 +361,21 @@ pipeline {
...
             withCredentials([aws(credentialsId: 'pmm-staging-slave')]) {
                 sh '''
                     set -o xtrace
+                    # read the files, not the env vars: those are only set once runSpotInstance
+                    # finishes, so a build aborted mid-launch would leave nothing to clean up
+                    REQ=$(cat REQUEST_ID 2>/dev/null)
+                    AMI=$(cat AMI_ID 2>/dev/null)
```
nogueiraanderson, line 371: When AMI_ID is missing, `AMI=$(cat AMI_ID 2>/dev/null)` exits this `sh` step, because Jenkins runs it with `-xe`. That is every no-capacity failure and every abort with a request still open. The cancel and the Slack notification never run, and an aborted build leaves its one-time request open for up to 7 days. Verified on a throwaway job. Let both reads return empty when the file is missing.

```suggestion
                    REQ=$(cat REQUEST_ID 2>/dev/null || true)
                    AMI=$(cat AMI_ID 2>/dev/null || true)
```

pmm/v3/pmm3-aws-staging-start.groovy
```diff
@@ -30,7 +30,7 @@ pipeline {
         )
         string(
             defaultValue: '3-dev-latest',
-            description: 'PMM Client version ("3-dev-latest" for main branch, "latest" or "X.X.X" for released version, "pmm3-rc" for Release Candidate, "http://..." for feature build)',
+            description: 'PMM Client version ("3-dev-latest" for main branch, "pmm3-latest" or "X.X.X" for released version, "pmm3-rc" for Release Candidate, "latest-tarball" or "https://....tar.gz" for feature build)',
```
nogueiraanderson, line 33: The host installs `pmm3-latest` from `experimental` (`setupPMM3Client`, line 75), while the containers get `release` from `install_pmm_client.yml`, so the description above is wrong for the host. In pmm-framework, `latest-tarball` resolves to the amd64 tarball, so containers on the arm64 VM get an x86_64 client until pmm-qa picks the architecture. Change line 75 of `setupPMM3Client` to `release` in this PR and mark `latest-tarball` as amd64 only.

```suggestion
            description: 'PMM Client version ("3-dev-latest" for main branch, "pmm3-latest" or "X.X.X" for released version, "pmm3-rc" for Release Candidate, "latest-tarball" (amd64 only) or "https://....tar.gz" for feature build)',
```

pmm/v3/pmm3-aws-staging-start.groovy
```diff
@@ -333,9 +329,10 @@ pipeline {
                         pushd /srv/pmm-qa/qa-integration/pmm_qa
                             echo "Setting docker based PMM clients"
 
+                            # pmm-framework needs this value as-is; changing it installs nothing
                             ./pmm-framework/pmm-framework \
                                 --pmm-server-password=${ADMIN_PASSWORD} \
-                                --client-version=${PMM_CLIENT_VERSION} \
+                                --client-version="${CLIENT_VERSION}" \
```
nogueiraanderson, line 335: Quoting `--client-version` preserves a trailing space that the old unquoted expansion dropped. pmm-framework then matches no install task and exits 0, the same silent no-op this PR fixes. The `setupPMM3Client` call already uses `.trim()`. Strip whitespace from the flag value as below.

```suggestion
                                --client-version="${CLIENT_VERSION//[[:space:]]/}" \
```

pmm/v3/vars/runSpotInstance.groovy
nogueiraanderson: Pinning every request to one AZ is the real limit. All three types (t4g.xlarge, m6g.xlarge, m7g.xlarge) are offered in every us-east-2 AZ today, so capacity explains the failures, not `constraint-not-fulfillable`. Live at 08:20 UTC, t4g.xlarge launched only in us-east-2c, and the combined placement scores were 9 in 2c, 5 in 2b, 1 in 2a. Replace lines 24 to 33 with the block below, then loop over `$SUBNETS` inside `for TYPE`, from before the request to after the cancel.

```sh
            SUBNETS=$(
                aws ec2 describe-subnets \
                    --region us-east-2 \
                    --output text \
                    --filters "Name=tag:aws:cloudformation:stack-name,Values=pmm-staging" \
                    --query 'Subnets[].SubnetId' \
                    | tr '\\t' '\\n' \
                    | sort --random-sort
            )
```

pmm/v3/vars/runSpotInstance.groovy
nogueiraanderson: When `request-spot-instances` fails, the script already exits a few lines later at `cancel-spot-instance-requests` with a MissingParameter error. Jenkins runs shebang-less `sh` steps as `sh -xe`, so errexit is active here, but `| tee REQUEST_ID` hides the aws exit status. A `continue` would turn a real API error into the no-capacity message. The cleaner fix is `pipefail` at the top of the block: a failed request then stops the build right there, with the real error as the last line, and no guard is needed. Tested: the subnet draw through `sort | head -1` stays safe under it. Replace `set -o xtrace` on line 9 with:

```sh
            set -o errexit
            set -o pipefail
            set -o xtrace
```

pmm/v3/vars/runSpotInstance.groovy
nogueiraanderson: An instance that launches just before its request is cancelled keeps running, and with the fan-out that can happen up to 15 times per build. The request ends in state `cancelled` with status `request-canceled-and-instance-running`, and `aws-staging-stop-robot` terminates running instances with a cancelled request every hour, tags or not, so the exposure is about an hour. To close it in the job, cancel first, then terminate whatever the request launched.

```suggestion
                        aws ec2 cancel-spot-instance-requests --region us-east-2 --spot-instance-request-ids $REQUEST_ID
                        LAUNCHED=$(
                            aws ec2 describe-instances \
                                --region us-east-2 \
                                --output text \
                                --filters "Name=spot-instance-request-id,Values=$REQUEST_ID" \
                                --query 'Reservations[].Instances[?State.Name!=`terminated`].InstanceId'
                        )
                        if [ -n "$LAUNCHED" ]; then
                            aws ec2 terminate-instances --region us-east-2 --instance-ids $LAUNCHED
                        fi
```

pmm/v3/vars/runSpotInstance.groovy
nogueiraanderson: Confirmed live: `SpotPriceHistory[0]` returned us-east-2a at 0.0527, so the round-1 cap is 0.0606. In 2b, m6g is 0.0687 and m7g 0.0719, so both sit price-too-low until round 3. The bid only sets a ceiling and the instance pays the market price. Delete this `SpotPrice` line to use the on-demand ceiling and make the rounds plain retries. If you want to keep the ramp, base it on the highest candidate price by replacing lines 55 to 62 with the block below.

```sh
                BASE_PRICE=$(
                    aws ec2 describe-spot-price-history \
                        --instance-types $CANDIDATE_TYPES \
                        --region us-east-2 \
                        --output text \
                        --product-description "Linux/UNIX (Amazon VPC)" \
                        --query 'SpotPriceHistory[].SpotPrice' \
                        | tr '\\t' '\\n' | sort -g | tail -1
                )
```

pmm/v3/vars/runSpotInstance.groovy
nogueiraanderson: The switch is right, the reason in the PR description is not. The robot copies the request's tags onto an untagged replacement and reaps it on the request `CreateTime`, so a persistent replacement was never invisible to it. What makes persistent useless here: the replacement boots the bare pmm-worker-3 AMI on a new IP with no PMM server and bills until the TTL, so it never brings back the owner's VM. Keep `one-time`, put that reason in the description, and update the DAYS text on line 58 as below.

```groovy
            description: 'Stop the instance in X days ("0" value disables the automated instance removal). On spot this is best effort and an interruption ends the VM earlier. Use USE_ONDEMAND to keep the VM for the whole period.',
```

pmm/v3/pmm3-aws-staging-start.groovy
nogueiraanderson: Talha's reading is right. `unsuccessful` covers ABORTED and `failure` does not, and the ssh wait loop after launch has no timeout, so aborts land there with a running instance to clean up. Verified on a throwaway job. Keep `unsuccessful` and make the reads on lines 370 to 371 tolerate a missing file, otherwise the post block exits before it cleans anything.

pmm/v3/vars/runSpotInstance.groovy
talhabinrizwan: @ademidoff, I don't think `persistent` is actually doing anything for us today.

We run `persistent + InstanceInterruptionBehavior: terminate`. That works exactly as configured, no errors, but it's the one pairing where `persistent` gives you nothing. `persistent` is meant to go with `stop/hibernate`: AWS stops the instance, preserves the EBS volume, and restarts the same instance when capacity returns. That's the "keep it for N days" behaviour you're protecting. With `terminate` the volume is destroyed, so there's nothing to restart, AWS just launches a blank replacement, new IP, no PMM server. The owner's environment is gone either way.

Live example: `sir-z79fhfmm`, from a build aborted on Sep 7. It's spawned 5 untagged instances, roughly one a day, and is still going. It escapes `aws-staging-stop-robot` because the request is untagged, `get_sir_days` returns `None`, so the age check is skipped entirely, and the request is still active rather than cancelled.

To make `persistent` genuinely useful we'd need `stop`, and that needs reaper changes first:
• `aws-staging-stop-robot.groovy` only queries `Values=running`, so stopped instances are invisible to it, 38 from October are still sitting there, `~$182/mo` in EBS
• `age` comes from `LaunchTime`, which resets on start, so the days clock restarts
• the public IP changes on stop/start, breaking the build-description link

Actual usage is 38/40 instances at days=1, 2 at days=5, none higher.

@nogueiraanderson, please correct me if I'm wrong.

pmm/v3/pmm3-aws-staging-start.groovy
talhabinrizwan: Went a different way on this one. Rather than aligning the host to release, I dropped `pmm3-latest` from the description entirely, nothing in the repo passes it (only occurrences are this description and the two comparisons in `setupPMM3Client`), and on the host it's byte-identical to `3-dev-latest`, both enable `experimental` and `dnf install pmm-client`. So it's an undocumented duplicate rather than a value worth fixing here, and `pmm3-latest → release` would land untested since nothing exercises it.

Took your `latest-tarball` note with one word changed, (docker clients amd64 only). The host is already arch-aware (setupPMM3Client:60-65 picks pmm-client-arm/… on aarch64), it's `normalize_client_version` in `pmm-qa` (pmm-framework/lib/common.sh:78) that hardcodes the amd64 URL. Same hole for X.X.X, install_pmm_client.yml:157,174 hardcode _amd64.deb/x86_64.rpm. Both go in the pmm-qa PR.

Also fixed `pmm3-deploy-services.groovy:123` in this PR, it advertised `latest` and feeds `CLIENT_VERSION` straight into this job..

pmm/v3/vars/runSpotInstance.groovy
talhabinrizwan: Done.

pmm/v3/vars/runSpotInstance.groovy
talhabinrizwan: Resolved by the rewrite: there is no bid any more. `SpotPrice` is gone entirely, so the cap defaults to on-demand and we pay market.

Your reading was right — derived from `t4g` alone, it was rejecting the very m6g/m7g fallbacks this PR added. @nogueiraanderson's measurements (t4g 0.0527, m6g 0.0687, m7g 0.0719 against a round-1 cap of 0.0606) are exactly why.

pmm/v3/vars/runSpotInstance.groovy
talhabinrizwan: Fixed. It now loops every `pmm-staging` subnet rather than drawing one, with an inner loop over each candidate type and `break 2` on the first success.

Verified in build 25656 with a bogus subnet prepended: all four types failed there, it moved to the next subnet and launched.

pmm/v3/vars/runSpotInstance.groovy
talhabinrizwan: Resolved by the rewrite: `request-spot-instances` and the `| tee REQUEST_ID` pipeline are both gone, so there is no masked exit status left to guard. `run-instances` failures are caught directly by the `if`.

Verified under `sh -xe` in 25656 (five consecutive failures, loop continued) and 25657 (all failed, clean `exit 1`).

pmm/v3/vars/runSpotInstance.groovy
talhabinrizwan: Resolved structurally. I applied your terminate-after-cancel block first (bd3affb9), but the `run-instances` rewrite then removed the cancel path entirely — the call is synchronous, so no request outlives it and nothing can launch behind our back.

Tags apply at launch now, so the untagged window is gone too.

pmm/v3/pmm3-aws-staging-start.groovy
talhabinrizwan: Applied.

Worth reporting back: build 25657 (exhaustion path, fault-injected) shows `cat REQUEST_ID` failing and `|| true` catching it. `REQUEST_ID` is only created after a successful launch, so on that path the file doesn't exist and the block would have aborted on its first line. Your catch was load-bearing.

pmm/v3/vars/runSpotInstance.groovy
```diff
@@ -17,151 +21,80 @@ def call(String INSTANCE_TYPE, boolean USE_ONDEMAND = false, String ARCH = 'x86_
...
+            fi
+
+            # tag at launch, so the VM is never running while invisible to aws-staging-stop
+            TAGS="[{Key=Name,Value=$VM_NAME},{Key=iit-billing-tag,Value=pmm-staging},{Key=stop-after-days,Value=$DAYS},{Key=owner,Value=$OWNER}]"
+
+            : > AMI_ID
+            for SUBNET in $SUBNETS; do
+                for TYPE in $CANDIDATE_TYPES; do
+                    # run-instances answers synchronously: either an id, or an error and no VM
+                    if aws ec2 run-instances \
```
nogueiraanderson, line 52: #### Every error is reported as no capacity

**Problem**

Any `run-instances` failure takes the same path: all nine subnet and type combinations are tried and the build ends with `Could not launch ... in any pmm-staging subnet`.

- Build 25657: `InvalidParameterValue` (arch mismatch) reported as capacity exhaustion.
- A bad key pair, a revoked credential, or a spot quota error reads the same way, and another AZ cannot fix any of them.
- `SUBNETS` comes from a pipeline without pipefail, so a denied `describe-subnets` gives zero attempts and the same message.

**Change**

Take the subnet list outside the pipeline so a failed `describe-subnets` aborts with its own error, guard the empty list, keep each attempt's stderr, and act on the exact error code:

```sh
SUBNET_LIST=$(
    aws ec2 describe-subnets \
        --region us-east-2 \
        --output text \
        --filters "Name=tag:aws:cloudformation:stack-name,Values=pmm-staging" \
        --query 'Subnets[].SubnetId'
)
SUBNETS=$(tr '\t' '\n' <<< "$SUBNET_LIST" | sort --random-sort)
[ -n "$SUBNETS" ] || { echo "No pmm-staging subnets found in us-east-2"; exit 1; }

: > AMI_ID
for SUBNET in $SUBNETS; do
    for TYPE in $CANDIDATE_TYPES; do
        if aws ec2 run-instances \
            --region us-east-2 ... --query 'Instances[].InstanceId' > AMI_ID 2> run.err; then
            break 2
        fi
        cat run.err
        ERROR_CODE=$(sed -nE 's/.*An error occurred \(([A-Za-z0-9.]+)\).*/\1/p' run.err | head -1)
        case "$ERROR_CODE" in
            InsufficientInstanceCapacity|Unsupported|SpotMaxPriceTooLow) ;;   # try the next type or subnet
            InsufficientFreeAddressesInSubnet) continue 2 ;;                  # this subnet is full, try the next one
            *) exit 1 ;;                                                      # not a capacity problem, stop here
        esac
    done
done
```

The exhaustion message then prints only when every combination was refused for capacity.

**Reproduce** (real AWS CLI v2 against a local EC2 fixture, no account or network, source in [probe-pr4441-launch-loop](https://github.com/nogueiraanderson/probe-pr4441-launch-loop)):

```sh
docker build -t pr4441-repro https://github.com/nogueiraanderson/probe-pr4441-launch-loop.git
docker run --rm --network none pr4441-repro
```

```
SCENARIO               VERSION CALLS HTTP EXIT LAST_MESSAGE
capacity-then-success  head    3     7    0    launched i-0123456789abcdef0 with IP 3.143.210.49
bad-keypair            head    9     9    1    Could not launch t4g.xlarge [tried t4g.xlarge m6g.xlarge m7g.xlarge] in any pmm-staging subnet
bad-keypair            fixed   1     1    1    An error occurred (InvalidKeyPair.NotFound) when calling the RunInstances operation: The key pair 'jenkins' does not exist
quota                  head    9     9    1    Could not launch t4g.xlarge [tried t4g.xlarge m6g.xlarge m7g.xlarge] in any pmm-staging subnet
quota                  fixed   1     1    1    An error occurred (MaxSpotInstanceCountExceeded) when calling the RunInstances operation: Max spot instance count exceeded
no-subnets             head    0     0    1    Could not launch t4g.xlarge [tried t4g.xlarge m6g.xlarge m7g.xlarge] in any pmm-staging subnet
no-subnets             fixed   0     0    254  An error occurred (UnauthorizedOperation) when calling the DescribeSubnets operation: You are not authorized to perform this operation.
```

CALLS is what the loop ran, HTTP is what the fixture received: the CLI retries a capacity refusal (HTTP 500) three times, which is the `(reached max retries: 2)` seen in production consoles, and does not retry client errors. The runner compares all 14 rows with the expected outcome and exits 1 on any difference.

pmm/v3/pmm3-aws-staging-start.groovy
```diff
@@ -30,7 +30,7 @@ pipeline {
         )
         string(
             defaultValue: '3-dev-latest',
-            description: 'PMM Client version ("3-dev-latest" for main branch, "latest" or "X.X.X" for released version, "pmm3-rc" for Release Candidate, "http://..." for feature build)',
+            description: 'PMM Client version ("3-dev-latest" for main branch, "X.X.X" for released version, "pmm3-rc" for Release Candidate, "latest-tarball" (docker clients amd64 only) or "https://....tar.gz" for feature build)',
```
nogueiraanderson: #### Three parameter descriptions to refresh

**This line, CLIENT_VERSION.** `X.X.X` also gives the docker clients an amd64 package (`install_pmm_client.yml` hardcodes the amd64 deb and the x86_64 rpm), the same limit as `latest-tarball`. Apply the same text at `pmm3-deploy-services.groovy:123`.

```suggestion
            description: 'PMM Client version ("3-dev-latest" for main branch, "X.X.X" for released version (docker clients amd64 only), "pmm3-rc" for Release Candidate, "latest-tarball" (docker clients amd64 only) or "https://....tar.gz" for feature build)',
```

**Line 28, SERVER_ARCH.** `arm64` can now land on `m6g.xlarge` or `m7g.xlarge` (my probe got m6g when t4g had no capacity in the AZ).

```groovy
            description: 'CPU architecture of the staging VM (arm64 = AWS Graviton, t4g/m6g/m7g.xlarge, first with capacity)',
```

**Line 58, DAYS.** On spot the VM is a one-time request with terminate on interruption, so DAYS is an upper bound.

```groovy
            description: 'Stop the instance in X days ("0" disables the automated removal). On spot this is best effort and an interruption ends the VM earlier. Use USE_ONDEMAND to keep the VM for the whole period.',
```

nogueiraanderson (commented):
#### Round 2 on fd191278

**Verdict:** merge after two small changes. Both round-1 blockers are fixed.

**Validated live** from the PR branch on pmm (throwaway job, every resource torn down):

- First-call `t4g.xlarge` spot launch, tags on instance and volume at launch.
- Real fallback: `t4g.xlarge` refused for capacity in one AZ, `m6g.xlarge` launched there 3 s later.
- On-demand path: tags at launch, no spot request.
- Abort during `wait instance-status-ok`: the post block terminated the VM in 2 s.

**Before merge:**

1. `pmm/v3/vars/runSpotInstance.groovy:48`: fail fast on non-capacity errors (snippet in the inline comment).
2. `pmm/v3/pmm3-aws-staging-start.groovy:33`: refresh three parameter descriptions (exact wording in the inline comment).

pmm/v3/vars/runSpotInstance.groovy
nogueiraanderson: Your reading is right, with one correction.

**Correction.** The robot's TTL check also uses the request `CreateTime`, which does not reset on stop and start, so only the `LaunchTime` leg restarts the clock.

**The leak is a loop now.** The untagged-instance cleanup terminates each replacement of `sir-z79fhfmm` within about 13 minutes and the persistent request launches the next one. One-time with tags at launch removes that class, so keep the switch.

**Left from this thread:**

- The DAYS wording on line 58 (text in my comment on line 33).
- Cancelling `sir-z79fhfmm` by hand. It is master's leak, this PR does not touch it.

kiveboxops:
@talhabinrizwan in general, it looks like it’s ready. Just a small fix I suggested for making some error messages more informative.

pmm/v3/vars/runSpotInstance.groovy
talhabinrizwan: Applied, with one addition: `InsufficientFreeAddressesInSubnet`. It's per-subnet and temporary, so the next AZ genuinely can work — but it isn't a capacity code, so under the original regex it would have stopped the build.

The `SUBNETS` guard is in too.

Verified both directions:

**Non-capacity now fails fast.** Build 25681, same `CANDIDATE_TYPES="t3.xlarge"` injection as 25657 — one `run-instances` attempt instead of three, the real error as the last line, `Could not launch` never printed, 22s instead of 80s.

```
+ cat run.err
aws: [ERROR]: An error occurred (InvalidParameterValue) … The architecture 'x86_64' of the
specified instance type does not match the architecture 'arm64' of the specified AMI
+ grep -qE 'InsufficientInstanceCapacity|…' run.err
+ exit 1
```

**Capacity still retries.** Build 25683 hit a genuine refusal, unprompted:

```
t4g.xlarge / subnet-75d2011d → InsufficientInstanceCapacity … in the Availability Zone
                               you requested (us-east-2a)
+ grep -qE … run.err          → matched, continued
m6g.xlarge / subnet-75d2011d → launched
```

That's the first real-world exercise of the fallback, and the build succeeded where master would have failed. `cat run.err` and the `grep` correctly don't run on a first-attempt success — `break 2` fires first.

pmm/v3/pmm3-aws-staging-start.groovy
talhabinrizwan: All three applied, plus the matching `X.X.X` note at `pmm3-deploy-services.groovy:123`.

You were right to push on `X.X.X` — I'd left it off deliberately as noise, but it's the same amd64 limit as `latest-tarball` (`install_pmm_client.yml:157,174`), so omitting it was just inconsistent.

`SERVER_ARCH` was outright wrong after this PR: 25683 landed on `m6g.xlarge` while the description still promised `t4g.xlarge`.

pmm/v3/vars/runSpotInstance.groovy
talhabinrizwan: Both items closed.

`DAYS` wording applied at line 58 — and it stopped being hypothetical a few hours later. Build 25683 launched an `m6g.xlarge` at 08:34 and AWS reclaimed it at 08:49:

```
sir-gfsfggsm  one-time  closed  instance-terminated-no-capacity
StateReason: Server.SpotInstanceTermination
```

Fifteen minutes against a requested `DAYS=1`, so the caveat earns its place.

It also shows the `one-time` switch behaving as you described: the request closed and nothing relaunched. Under `persistent` that would have been a bare `pmm-worker-3` AMI on a new IP, billing until the TTL. No orphaned volume either — the tagged EBS went with the instance.

`sir-z79fhfmm` was already cancelled by the time I looked (`instance-terminated-by-user`, instance `i-06cca044cf371aa4c` terminated). I checked for siblings and found only `sir-2iqfjgvq`, which is a healthy tagged nightly VM, not a leak — left alone.

kiveboxops (dismissed):

nogueiraanderson (approved):
Verified head b26fdfc2: the three descriptions, the SUBNETS guard, and the fail-fast path are in, and build 25683 shows the fallback and a real spot reclaim behaving as designed. Approving.

One optional nit: anchor the capacity match to the error code, `grep -qE 'An error occurred \((InsufficientInstanceCapacity|InsufficientFreeAddressesInSubnet|SpotMaxPriceTooLow|Unsupported)\)' run.err`, so `UnsupportedOperation` cannot be swept and the quota codes fail fast too. Alex's request is the remaining gate.

pmm/v3/vars/runSpotInstance.groovy
```diff
@@ -17,151 +21,88 @@ def call(String INSTANCE_TYPE, boolean USE_ONDEMAND = false, String ARCH = 'x86_
...
+                        --query 'Instances[].InstanceId' > AMI_ID 2> run.err
+                    then
+                        break 2
                     fi
 
-                    aws ec2 cancel-spot-instance-requests --region us-east-2 --spot-instance-request-ids $REQUEST_ID
-                    PRICE_MULTIPLIER=$((PRICE_MULTIPLIER+1))
+                    # only a capacity refusal is worth another subnet or type
+                    cat run.err
+                    grep -qE 'InsufficientInstanceCapacity|InsufficientFreeAddressesInSubnet|SpotMaxPriceTooLow|MaxSpotInstanceCountExceeded|InstanceLimitExceeded|Unsupported' run.err || exit 1
```
nogueiraanderson, line 71: One row differs from the round-2 table: `MaxSpotInstanceCountExceeded` and `InstanceLimitExceeded` are in the retry pattern, so a quota refusal runs all nine attempts and ends with `Could not launch ... in any pmm-staging subnet`. Both quotas are per region and per family (t4g, m6g and m7g share the Standard family), so another AZ or type cannot clear them.

Fixture, this block verbatim, real CLI:

```
SCENARIO  VERSION   CALLS HTTP EXIT LAST_MESSAGE
quota     proposed  1     1    1    An error occurred (MaxSpotInstanceCountExceeded) when calling the RunInstances operation
quota     b26fdfc2  9     9    1    Could not launch t4g.xlarge [tried t4g.xlarge m6g.xlarge m7g.xlarge] in any pmm-staging subnet
```

Not blocking: the real error is printed before each retry and nothing is left running. For the fast exit, drop the two codes.

```suggestion
                    grep -qE 'InsufficientInstanceCapacity|InsufficientFreeAddressesInSubnet|SpotMaxPriceTooLow|Unsupported' run.err || exit 1
```

nogueiraanderson (approved):
#### Round 3 on b26fdfc2

**Verdict:** approve. Both round-2 items are in: the three descriptions (9451c5bc) as proposed, and the fail-fast with the subnet guard (b26fdfc2).

**Checked** with the round-2 fixture, extended with the b26fdfc2 block verbatim as a third variant and the same expectations as the proposed change (real CLI 2.36.44):

| scenario | b26fdfc2 |
|---|---|
| success, capacity-then-success, all-capacity | as proposed |
| bad-keypair | 1 attempt, real error, exit 1 |
| subnet-full | next attempt launches |
| no-subnets | 0 attempts, `No pmm-staging subnets found in us-east-2`, DescribeSubnets error one line above |
| quota | 9 attempts, ends with the capacity message (inline note on line 71, not blocking) |

Builds 25681 and 25683 cover the live side, including the first real m6g fallback and the spot reclaim the DAYS text now warns about.

The changes-requested review from f4e177e still gates the merge. Each point in it is answered in the threads or removed by the rewrite, so it needs @ademidoff's re-review.

yurkovychv (approved):

pmm/v3/vars/runSpotInstance.groovy
```diff
@@ -17,151 +21,88 @@ def call(String INSTANCE_TYPE, boolean USE_ONDEMAND = false, String ARCH = 'x86_
...
-                AMI_ID=$(
-                    aws ec2 run-instances \
+            if [ -z "$SUBNETS" ]; then
+                echo "No pmm-staging subnets found in us-east-2"
+                exit 1
+            fi
+
+            MARKET_OPTS=""
+            if [ "$USE_ONDEMAND" != "true" ]; then
+                MARKET_OPTS="--instance-market-options MarketType=spot"
```
ademidoff: `run-instances` defaults `SpotOptions.SpotInstanceType` to `one-time`, so this is a one-time request — not the `persistent` behaviour we settled on. Needs spelling out:

```
MARKET_OPTS="--instance-market-options MarketType=spot,SpotOptions={SpotInstanceType=persistent}"
```

(dry-run against the pmm-worker-3 AMI and a pmm-staging subnet accepts it.)

pmm/v3/vars/runSpotInstance.groovy
```diff
@@ -17,151 +21,88 @@ def call(String INSTANCE_TYPE, boolean USE_ONDEMAND = false, String ARCH = 'x86_
...
                         --image-id "$IMAGE_ID" \
-                        --instance-type "$INSTANCE_TYPE" \
+                        --instance-type "$TYPE" \
                         --key-name jenkins \
                         --iam-instance-profile Name=pmm-staging-slave \
                         --security-group-ids sg-cd39dba6 sg-9f3cdef4 sg-0cbb55499c1e70fb7 \
                         --subnet-id "$SUBNET" \
                         --count 1 \
+                        $MARKET_OPTS \
+                        --tag-specifications "ResourceType=instance,Tags=$TAGS" "ResourceType=volume,Tags=$TAGS" \
```
ademidoff: Going `persistent` needs the request tagged too: `"ResourceType=spot-instances-request,Tags=$TAGS"`. A replacement instance the request launches isn't covered by the instance tag spec, and `pmm/aws-staging-stop-robot.groovy:139-142` recovers those by copying tags off the spot request — an untagged request means an untagged replacement the reaper never sees.

pmm/v3/pmm3-aws-staging-start.groovy
```diff
@@ -55,7 +55,7 @@ pipeline {
         )
         choice(
             choices: '1\n0\n1\n2\n3\n4\n5\n6\n7\n8\n9\n10\n11\n12\n13\n14\n15\n16\n17\n18\n19\n20\n21\n22\n23\n24\n25\n26\n27\n28\n29\n30',
-            description: 'Stop the instance in X days ("0" value disables the automated instance removal)',
+            description: 'Stop the instance in X days ("0" disables the automated removal). On spot this is best effort and an interruption ends the VM earlier. Use USE_ONDEMAND to keep the VM for the whole period.',
```
ademidoff: Stale once the request is `persistent`: an interruption no longer ends the VM, the request stays open and a replacement launches. The "best effort / use USE_ONDEMAND" half doesn't hold.

ademidoff (approved):

PR #4463, opened 2026-09-17
PMM-7 Unblock the nightly: client tarball and junit path

travagliad:
Two independent blockers on the nightly, one commit each.

---

# 1 — Default the nightly main lanes to the client tarball

One file, two lines: the `CLIENT_VERSION` default of `pmm3-nightly-orchestrator` moves from `3-dev-latest` to `latest-tarball`.

## Why

`3-dev-latest` installs the PMM client as a **package from `repo.percona.com`**. In orchestrator build #12 that is what broke the arm64 lane. From its `setup / mysql-psmdb-pgsql` log:

```
TASK [Fetch and verify the pmm-client package on the host]
fatal: [localhost]: FAILED! => .../fetch-pmm-client-deb.sh experimental ...
  [fetch-pmm-client-deb] download failed (attempt 1)
  [fetch-pmm-client-deb] giving up after 600s (last failure: download-failed).
  curl: (28) Operation timed out after 900000 milliseconds
        with 169693171 out of 180482458 bytes received
Fetch and verify the pmm-client package on the host ------- 900.46s
```

A 180.5 MB deb that stalled at ~170 MB and timed out, twice, 900.46 s each — this is the `ERROR: mysql/mysql-setup.yml playbook execution failed` that took the lane down. The pgsql container in the same run spent 15m40s and 16m37s in `Install pmm-client`, via `pmm3-client-setup.sh`, whose repo path retries five times ninety seconds apart. `repo.percona.com` appears 31 times in that one setup job.

## What this is *not*

It is not a bandwidth optimisation, and it should not be expected to speed up a healthy setup:

| | size | throughput, measured side by side outside CI |
| --- | --- | --- |
| deb from `repo.percona.com` | 180.5 MB | 25.7 MB/s |
| tarball from S3 | 187.0 MB | 29.4 MB/s |

Same bytes, comparable rates. What changes is the tail. The packages are built *from* the tarball, so the tarball is upstream of them, and a client build running concurrently holds the repo until it has finished pushing — which is the window that timeout landed in. Moving to S3 leaves that window and replaces apt's index dance plus the five-by-ninety-second retry with a single request. The repo is still reached for `percona-release` itself (19.4 KB) on every Debian container, so this is not "no repo traffic".

## Blast radius

`pmm3-ui-tests-nightly-gha` already defaults to `latest-tarball`; the orchestrator was overriding it. Dropping the override moves:

- `nightly / docker`, `docker arm64`, `ami`, `helm`, `ha` (via the `nightlyGha` defaults map)
- `ui / @ia`, `@instances`, `@gcp`

and leaves everything else untouched, because each of those passes `CLIENT_VERSION` explicitly:

- `compat / client X.Y.Z` — overrides with the GA version under test
- `pkg amd64` / `pkg arm64` — `packageBranches` never takes `CLIENT_VERSION`; package testing is meant to exercise the repo
- `upgrade / *` — already computes a versioned tarball URL, falling back to a deb only while one is still in the apt pool
- `nightly / gssapi` — already passes a tarball URL

## Verification

- **Groovy**: `CompilationUnit` parse to `Phases.CONVERSION` — OK.
- **Parameter merge**: executed the `(defaults + cfg)` logic — docker/arm64/helm resolve to `latest-tarball`, the `compat 3.7.1` override still wins.
- **Ansible (core 2.19.13)**, against `install_pmm_client.yml`'s own gate and branch conditions:
  - the `assert` at the top of that file accepts `3-dev-latest`, `3.7.1` and a tarball URL, and rejects the bare alias `latest-tarball` — which is fine, because every entry point expands it first: all 15 client-version assignments in `pmm-framework/setups/*.sh` go through `resolved_client_version` → `normalize_client_version`, and `pmm3-client-setup.sh` expands it itself.
  - branch selection with a tarball URL: `pmm_client_component` is empty, the tarball task fires, both package tasks skip.
  - branch selection with `3.7.1`: `component=main`, the Debian package path still fires — so the compat lanes are genuinely unaffected.
- **Tarball URLs live**: x86 and arm64 both serve (`200` / `206`).

## Residual risk

The three `ui /` lanes get `latest-tarball` for the first time — `pmm3-ui-tests`'s own default is `3-dev-latest`. `setupPMM3Client` handles the alias (it even picks the arm64 URL by `uname -m`), and the one suspicious line, `pmm3-ui-tests.groovy:336` prepending `pmm-client/bin` to `PATH` when the version is not `3-dev-latest`, is inert if that directory does not exist. That is reasoning, not a run. If you would rather isolate the variable on the first nightly after this merges, I can pin the `ui /` lanes to `3-dev-latest` explicitly so only the five `nightly /` lanes move.

## Context

Raised by @yurkovychv in today's QA sync: "Please switch to latest-tarball, because you're actually blowing up the repo, and that's why we have latest-tarball by default everywhere… that's why we need to use tarballs everywhere where possible by default." He explicitly exempted package tests, and left the upgrade pipeline as "we can think about it" — both are untouched here.

You can try this before it merges, without this branch: the orchestrator takes `CLIENT_VERSION` as a build parameter, so triggering it with `latest-tarball` typed in gives the same behaviour.

---

# 2 — Drop the junit path `pmm3-upgrade-test-runner` no longer writes to

**`pmm3-upgrade-test-runner` cannot currently go green, whatever the tests do.** The last thirty builds are `FAILURE` or `ABORTED` — not one `SUCCESS` — and that includes runs whose every suite reported `0 failed`.

The `post { always }` block ran two `junit` steps:

```groovy
def PATH_TO_REPORT_RESULTS = 'tests/output/*.xml'
try {
    dir('/home/ec2-user/workspace/pmm3-upgrade-test-runner') {   // nothing writes here any more
        junit PATH_TO_REPORT_RESULTS
    }
} catch (err) {
    error "No test reports found at path: " + PATH_TO_REPORT_RESULTS   // aborts the block
}
try {
    dir('/srv/pmm-qa/codeceptjs-e2e') {                          // the real reports — never reached
        junit PATH_TO_REPORT_RESULTS
    }
} catch (err) { ... }
```

#4461 moved the codeceptjs runs out of the Jenkins workspace and into `/srv/pmm-qa/codeceptjs-e2e`, but left the first step behind. It now finds no XML, its `catch` calls `error()`, and that aborts the block before the second step — the one pointed at the reports that do exist — can run.

`git log -S` dates both: the `/srv/pmm-qa/codeceptjs-e2e` step has been there since 05b9b1d; the workspace step came with 5f2b4d0, when the pre-upgrade suite still ran in the workspace against a `pmm-ui-tests` checkout. Since 6a6ba7b it has had nothing to read.

This commit removes the workspace `junit` step and the matching failure-time screenshot archive, both of which predate the move.

### Evidence

Build [23690](https://pmm.cd.percona.com/job/pmm3-upgrade-test-runner/23690/) (`SSL`), run on a pmm-qa branch that fixes the suite-side breakage from the same move:

```
OK | 3 passed,  9 skipped   // pre-ssl-upgrade
OK | 9 passed,  3 skipped   // post-server-upgrade
OK | 9 passed,  3 skipped   // post-client-upgrade
ERROR: No test reports found at path: tests/output/*.xml
→ FAILURE
```

### Not in scope here

`Prepare` still clones `percona/pmm-ui-tests` into the Jenkins workspace and `Install dependencies` still runs `npm ci`, `envsubst < env.list` and `sed … pr.codecept.js` there. Nothing downstream consumes any of it — `env.generated.list` is referenced nowhere else in the pipeline — so that checkout looks dead too, but removing it is a separate change and is not what turns a passing run red.

### Verification

`groovyc` compiles the edited pipeline cleanly (`rc=0`).

---

# 3 — The rest of the branch, and why it is one branch

This branch has grown past the two commits above, because the goal is that **every
`pmm3-*` job can point at this one ref and get every fix at once** — one merge, one
branch to configure, one orchestrator run that exercises all of it.

## Own commits beyond the two above

| Commit | What |
| --- | --- |
| `8a9460cc` | Finish the move off the `pmm-ui-tests` checkout in `pmm3-upgrade-ami-test-runner` (the leftover `Not in scope here` from §2) |
| `83d6fafa` | Clone `pmm-qa` shallow (`depth: 1`) on the lanes the nightly fans out |
| `78049ec2` | Three upgrade-runner failures seen in nightly #18 |
| `3220a801` | Pre-pull the upgrade image on the UI upgrade path too |
| `0e1b359d` | Default the nightly to the Docker upgrade method |
| `bb030490` | `--setup-retries 1` on the upgrade runner, so one failed database setup does not take the lane down |

## Cherry-picked from the open PMM-15197 branches

So that this branch is the only one that needs merging and configuring, the
on-demand-agent work from `PMM-7-fix-ami-upgrade` and
`PMM-7-fix-nightly-gssapi-3_8_0` is carried here as `-x` cherry-picks, authorship
intact:

- `62c3f201` — `PMM-15197 Run the AMI upgrade tests on on-demand agents`
- `8b612f93` — `PMM-15197 Give the AMI upgrade runner 90 minutes and retry client installs (#4438)`
- `29b89c4e` — `PMM-15197 Run the GSSAPI nightly on on-demand agents`
- `cdeaf01a` — `PMM-15197 Pass USE_ONDEMAND to the staging server jobs`

A straight branch merge was tried first and rejected: both branches fork from before
`pmm3-upgrade-ami-test-runner` moved from `pmm-ui-tests` to `pmm-qa`, so merging them
resurrects `PMM_UI_PRE_UPGRADE_GIT_BRANCH`, the `pmm-ui-tests.git` checkout, the
workspace-root `npm ci`, and drops the `SSH_KEY` parameter. The cherry-picks take only
the four intended changes — the `USE_ONDEMAND` parameter and the conditional agent
labels in `pmm3-upgrade-ami-test-runner`, `pmm3-upgrade-ami-tests` and
`pmm3-ui-tests-nightly-gssapi`, plus the `install_client` retry loop — and leave the
`pmm-qa` migration intact. `SSH_KEY` and `USE_ONDEMAND` now both exist on the runner.

### Verification

- `CompilationUnit` parse to `Phases.CONVERSION` on all three touched files — OK.
- Every job named in a new `USE_ONDEMAND` passthrough declares that parameter:
  `pmm3-aws-staging-start`, `pmm/aws-staging-stop`, `pmm3-ami-staging-stop`,
  `pmm3-upgrade-ami-test-runner`.
- Branch is level with `master` and merges cleanly.

https://claude.ai/code/session_01XaD6E1PsySaavZhpASfizu

---
_Generated by [Claude Code](https://claude.ai/code)_


pmm/v3/pmm3-nightly-orchestrator.groovy
```diff
@@ -268,13 +267,17 @@ def amiUpgradeBranches(Map branches, String serverImage, String latestDevVersion
     pmmVersion('v3')[-5..-1].each { ver ->
         def name = "upgrade / ami ${ver}"
         branches[name] = suite(name, 'pmm3-upgrade-ami-test-runner', [
-            string(name: 'PMM_UI_PRE_UPGRADE_GIT_BRANCH', value: "pmm-${ver}"),
             string(name: 'PMM_QA_GIT_BRANCH',             value: params.PMM_QA_GIT_BRANCH),
+            // Without this the runner falls back to its own default, which is pinned to
+            // the oldest version, so every lane ran 3.7.1's pre-upgrade suite against the
+            // version it was actually upgrading from.
```
```suggestion
```

pmm/v3/pmm3-ui-tests-nightly-gha.groovy
```diff
@@ -280,11 +285,16 @@ pipeline {
 
                         chmod +x .github/scripts/wait-for-gh-run.sh .github/scripts/wait-for-gh-run-completion.sh
 
+                        # VM_IP is what the workflow puts at the end of its run-name, and
+                        # it is the only thing distinguishing this lane's dispatch from the
+                        # nine others that hit the same file on the same branch seconds
+                        # either side of it.
```
travagliad:
```suggestion
```

pmm/v3/pmm3-ui-tests-nightly-gha.groovy
```diff
@@ -305,12 +315,14 @@ pipeline {
                         string(name: 'AMI_ID', value: env.AMI_INSTANCE_ID),
                     ]
                 }
+                // Not FORCE_MODE: without an INFRA_ID it destroys by the bare cluster
+                // name, which owns no AWS resources, and reports success having removed
+                // nothing. The create job left the state in S3, so let destroy use it.
```
travagliad:
```suggestion
```

pmm/v3/pmm3-ui-tests-nightly-gssapi.groovy
```diff
@@ -71,7 +74,10 @@ void checkClientNodesAgentStatus(String VM_CLIENT_IP, PMM_QA_GIT_BRANCH) {
                 set -o xtrace
                 echo "Checking Agent Status on Client Nodes";
                 sudo mkdir -p /srv/pmm-qa || :
-                sudo git clone --single-branch --branch $PMM_QA_GIT_BRANCH https://github.com/percona/pmm-qa.git /srv/pmm-qa
+                # pmm3-aws-staging-start already clones pmm-qa here at this same branch, and
+                # git refuses to clone into a non-empty directory. Under errexit that fatal
+                # ends the whole nightly with exit 128, so treat an existing checkout as fine.
```
travagliad:
```suggestion
```

pmm/v3/pmm3-upgrade-ami-test-runner.groovy
```diff
@@ -153,14 +161,22 @@ pipeline {
...
+                    branches: [[name: PMM_QA_PRE_UPGRADE_GIT_BRANCH]],
+                    userRemoteConfigs: [[url: 'https://github.com/percona/pmm-qa.git']],
+                    extensions: [[$class: 'CloneOption', shallow: true, depth: 1]],
+                ]
 
+                // The workspace stays on the pre-upgrade branch for the pre-upgrade suite,
+                // but /srv/pmm-qa supplies tooling -- pmm-framework and support_scripts --
+                // which a release branch that old does not carry: pmm-3.7.1 has no
+                // pmm-framework at all, so rsyncing the workspace here left every AMI lane
+                // failing on "./pmm-framework/pmm-framework: No such file or directory".
```
travagliad:
```suggestion
```

pmm/v3/pmm3-upgrade-ami-test-runner.groovy
```diff
@@ -248,36 +264,86 @@ pipeline {
...
+                    }
+                    steps {
+                        withCredentials([aws(accessKeyVariable: 'BACKUP_LOCATION_ACCESS_KEY', credentialsId: 'BACKUP_E2E_TESTS', secretKeyVariable: 'BACKUP_LOCATION_SECRET_KEY'), aws(accessKeyVariable: 'AWS_ACCESS_KEY_ID', credentialsId: 'PMM_AWS_DEV', secretKeyVariable: 'AWS_SECRET_ACCESS_KEY')]) {
+                            dir('codeceptjs-e2e') {
+                                // @ami-upgrade lived in pmm-ui-tests and was never carried into
+                                // pmm-qa, so this stage matched nothing since the migration.
+                                // @pmm-upgrade is the suite that drives the in-app update; PMM 3.9
+                                // removed that button ("All updates are now securely managed via
+                                // the CLI"), so the orchestrator sends 3.9+ lanes down the DOCKER
+                                // branch below, the same gate it applies to the docker runner.
```
travagliad:
```suggestion
```

pmm/v3/pmm3-upgrade-ami-test-runner.groovy
```diff
@@ -293,10 +359,14 @@ pipeline {
                         "'
                     """
                 }
+                // The UI reports the update done as soon as the new container is up; Grafana
+                // behind it still needs time, and the first post-upgrade page load only gets
+                // 60 seconds for the shell to appear before the suite calls it a failure.
```
travagliad:
```suggestion
```

pmm/v3/pmm3-upgrade-ami-test-runner.groovy
```diff
@@ -374,7 +456,10 @@ pipeline {
             }
         }
         failure {
-            archiveArtifacts artifacts: 'codeceptjs-e2e/tests/output/parallel_chunk*/*.png'
+            // Plain `codeceptjs run` writes screenshots straight into tests/output,
+            // not into the parallel_chunk* directories run-multiple created, and an
+            // archiveArtifacts that matches nothing fails the build on its own.
```
travagliad:
```suggestion
```

pmm/v3/pmm3-upgrade-test-runner.groovy
```diff
@@ -371,9 +358,23 @@ pipeline {
                         expression { return params.UPGRADE_TYPE == "UI" }
                     }
                     steps {
+                        // Watchtower pulls this image itself once the UI asks it to, inside the
+                        // 240 seconds PMM-T3 allows for the whole upgrade. The Docker path
+                        // already pulls it up front with the same retry loop; doing it here too
+                        // leaves watchtower only the container swap, so a stall can no longer be
+                        // the download.
```
travagliad:
```suggestion
```

pmm/v3/pmm3-upgrade-test-runner.groovy
```diff
@@ -442,6 +443,21 @@ pipeline {
             steps {
                 withCredentials([aws(accessKeyVariable: 'BACKUP_LOCATION_ACCESS_KEY', credentialsId: 'BACKUP_E2E_TESTS', secretKeyVariable: 'BACKUP_LOCATION_SECRET_KEY'), aws(accessKeyVariable: 'AWS_ACCESS_KEY_ID', credentialsId: 'PMM_AWS_DEV', secretKeyVariable: 'AWS_SECRET_ACCESS_KEY')]) {
                     sh '''
+                        # A container whose client was installed from a tarball has no
+                        # percona-release, and an empty CLIENT_TARBALL_UPGRADE still routes it
+                        # down the package path -- external_pmm then exits 127 on the whole stage.
```
travagliad:
```suggestion
```

pmm/v3/pmm3-upgrade-test-runner.groovy
```diff
@@ -631,6 +654,12 @@ pipeline {
                 docker exec pmm-server cat /srv/logs/pmm-managed.log >> pmm-managed-full.log || true
                 docker exec pmm-server cat /srv/logs/pmm-update-perform.log >> pmm-update-perform.log || true
                 echo --- pmm-update-perform logs from pmm-server --- >> pmm-update-perform.log
+
+                # pmm-managed only hands the update to watchtower, which then pulls the image
+                # and recreates the container; nothing else records what watchtower did, so a
+                # stalled upgrade shows up as "Successfully triggered update" and silence.
```
travagliad:
```suggestion
```

pmm/aws-staging-stop.groovy
```diff
@@ -92,7 +90,7 @@ pipeline {
                             exit 1
                         fi
                         # On-demand instances have no spot request -- AWS renders that as the
-                        # literal "None" in the table above, not an empty string.
+                        # literal "None", not an empty string.
```
travagliad:
```suggestion
                        # literal "None" in the table above, not an empty string.
```

pmm/aws-staging-stop.groovy
```diff
@@ -78,11 +78,9 @@ pipeline {
                     sh '''
                         set -o errexit
                         set +x
-                        # cut on '|' instead of awk fields -- awk's whitespace splitting
-                        # breaks when a short value like "None" doesn't butt up against the next '|'.
                         ROW=$(echo "${VMList}" | grep "${INPUT}")
```
travagliad, line 83:
```suggestion
                        # cut on '|' instead of awk fields -- awk's whitespace splitting
                        # breaks when a short value like "None" doesn't butt up against the next '|'.
                        ROW=$(echo "${VMList}" | grep "${INPUT}")
```

travagliad: Not restoring these two, and here is why — flagging it rather than quietly dropping them.

Both comments are on `master` and both describe the code the leak fix replaces. The `aws ec2 describe-instances` above now runs with `--output text` instead of `--output table`:

- `# cut on '|' instead of awk fields -- awk's whitespace splitting breaks when a short value like "None" doesn't butt up against the next '|'` justifies `cut -d '|' -f2`. That was the bug: with one matching row `--output table` switches to a vertical key/value layout, so field 2 is not the request id. Restored verbatim it would sit directly above two `awk '{print $1}'` / `awk '{print $2}'` lines and say the opposite of what the code does. `--output text` is tab-separated in key order, which is why `awk` is now correct — and it is the same shape `aws-staging-stop-robot` already uses.
- `# literal "None" in the table above` — "the table" is `--output table`, which is gone. The line kept on the branch is the same statement with the stale reference removed: `# literal "None", not an empty string.`

If the intent was that a shared `master` job's comments shouldn't be touched in this PR at all, say so and I'll pull the whole `aws-staging-stop` change (`9cfc7e13`, `9738cb74`, `2478ec4d`) out into its own PR instead — the comments and the code have to agree either way.

---
_Generated by [Claude Code](https://claude.ai/code)_

pmm/v3/pmm3-nightly-orchestrator.groovy
```diff
@@ -401,25 +412,55 @@ timestamps {
                                 skip_compatibility     : false,
                             ],
                         ]).toString()
-                        withCredentials([string(credentialsId: 'GITHUB_API_TOKEN', variable: 'GITHUB_TOKEN')]) {
+                        // Dispatching and walking away recorded DISPATCHED for a run
+                        // nobody then read. The orchestrator is the only thing that
+                        // sees the whole night, so this lane waits for its verdict
+                        // like every other one.
```
```suggestion
```

pmm/v3/pmm3-ami.groovy
```diff
@@ -61,6 +63,24 @@ pipeline {
...
+            script {
+                // The nightly orchestrator tests this AMI instead of the last GA
+                // one, so it is chained here rather than kept on a cron of its
+                // own. A release candidate does not chain here: pmm3-release-candidate
+                // triggers the orchestrator itself, with the AMI this build returns.
+                // Nothing else gates the handoff -- whatever AMI this build
+                // produced is passed on, and a failed or aborted build passes
+                // none, which fails the orchestrator's AMI lane on an empty id
+                // rather than quietly testing an older image while the docker,
+                // helm, package and upgrade suites run.
```
travagliad:
```suggestion
```

talhabinrizwan (approved):

